PR #5, opened 2017-03-07
Resolve race condition in Server KEX Algorithm fields

bvandersloot:
Fixes a race condition that arises when using multiple permitted
server key exchange algorithms in the same process

Depends upon #4 to pass tests.


bvandersloot:
Merging #4, as promised, fixed breaking tests.

PR #6, opened 2017-03-07
Fix nil pointer dereference in big.Int.Cmp().

aaspring:


tls/tls_ka.go
```diff
@@ -102,9 +102,9 @@ func (ka *dheKeyAgreement) DHParams() *keys.DHParams {
 	}
 	if ka.yServer != nil {
 		out.ServerPublic = new(big.Int).Set(ka.yServer)
-	}
-	if ka.yServer.Cmp(ka.yOurs) == 0 && ka.xOurs != nil {
-		out.ServerPrivate = new(big.Int).Set(ka.xOurs)
+		if ka.yOurs != nil && ka.yServer.Cmp(ka.yOurs) == 0 && ka.xOurs != nil {
```
zakird: Shouldn't we check both ` ka.yOurs != nil` and `ka.xOurs != nil` before checking `ka.yServer.Cmp(ka.yOurs)`?

zakird: Also this indentation looks incorrect.

tls/tls_ka.go
```diff
@@ -119,9 +119,9 @@ func (ka *dheKeyAgreement) ClientDHParams() *keys.DHParams {
 	}
 	if ka.yClient != nil {
 		out.ClientPublic = new(big.Int).Set(ka.yClient)
-	}
-	if ka.yClient.Cmp(ka.yOurs) == 0 && ka.xOurs != nil {
-		out.ClientPrivate = new(big.Int).Set(ka.xOurs)
+		if ka.yOurs != nil && ka.yClient.Cmp(ka.yOurs) == 0 && ka.xOurs != nil {
+			out.ClientPrivate = new(big.Int).Set(ka.xOurs)
+		}
```
zakird: Same issues here as above.

PR #16, opened 2017-03-15
Initial framing for multi browser validation

dadrian:


verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+// VerificationResult contains the result of a verification of a certificate
+// against a browser. It discerns between NameError, meaning the name on the
+// certificate does not match, and a ValidationError, meaning some sort of issue
+// with the cryptography, or other issues with the chain.
+type VerificationResult struct {
+
+	// The name being checked, e.g. "www.censys.io".
+	Name string
```
zakird: Why is this returned?

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+	// built.
+	ValidationError error
+
+	// NameError will be non-nil when there was a mismatch between the name on the
+	// certificate and the name being verified against.
+	NameError error
+
+	// Chains is a list of validated certificate chains, starting at the
+	// certificate being verified, and ending at a certificate in the root store.
+	Chains [][]*x509.Certificate
```
zakird: Why are we returning multiple chains?

zakird: Why aren't we returning any information about whether the certificate chained up to the root store?

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+// with the cryptography, or other issues with the chain.
+type VerificationResult struct {
+
+	// The name being checked, e.g. "www.censys.io".
+	Name string
+
+	// True if the certificate is whitelisted by a given browser as trusted, e.g.
+	// when Apple whitelisted subset of StartCom certs by SPKI hash in the wake of
+	// the WoSign incidents surrounding misissuance.
+	Whitelisted bool
```
zakird: Why are we returning this? It doesn't make sense as a return result of this level call? What happens for example if something is both whitelisted and blacklisted?

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+
+// MatchesDomain returns true if NameError == nil and Name != "".
+func (res *VerificationResult) MatchesDomain() bool {
+	return res.NameError == nil && res.Name != ""
+}
+
+// VerifyProcedure is an interface to implement additional browser specific logic at
+// the start and end of verification.
+type VerifyProcedure interface {
+	PreValidate(c *x509.Certificate) error
```
zakird: Why doesn't this take a chain in?

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+// A Verifier represents a context for verifying certificates.
+type Verifier struct {
+	Roots           *x509.CertPool
+	Intermediates   *x509.CertPool
+	Blacklist       *x509.CertPool
+	Whitelist       *x509.CertPool
+	VerifyProcedure VerifyProcedure
+}
+
+func (v *Verifier) convertOptions(opt *VerificationOptions) (out x509.VerifyOptions) {
```
zakird: Why are we implementing this?

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+	if opt.VerifyTime.IsZero() {
+		opt.VerifyTime = time.Now()
+	}
+}
+
+// A Verifier represents a context for verifying certificates.
+type Verifier struct {
+	Roots           *x509.CertPool
+	Intermediates   *x509.CertPool
+	Blacklist       *x509.CertPool
```
zakird: I'm not sure this is the right way of storing this. I'm guessing that browsers aren't shipping entire certificates around, and I think we're getting ahead of ourselves.

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+	out.DNSName = opt.Name
+	return
+}
+
+// Verify checks if c chains back to a certificate in Roots, possibly via a
+// certificate in Intermediates, and returns all such chains. It additional
+// checks if the Name in the VerificationOptions matches the name on the
+// certificate. Finally, it checks to see if c is blacklisted or whitelisted by
+// the Browser.
+func (b *Browser) Verify(c *x509.Certificate, opts VerificationOptions) (res *VerificationResult) {
```
zakird: This doesn't seem like the right thing to call this, and we're already doing too much in this function. This should probably just become `GetChainsToRoot` or similar

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// Run PostValidate if there was no exisiting validation error.
+	if res.ValidationError == nil {
+		if err := v.VerifyProcedure.PostValidate(c, res.Chains); err != nil {
+			res.ValidationError = err
+		}
+	}
+
+	// Check the whitelist
+	if v.Whitelist.Contains(c) {
```
zakird: Again, we have _no_ idea if this is the right way of doing this, and right now isn't the time to be working this out.

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+
+import (
+	"time"
+
+	"github.com/zmap/zcrypto/x509"
+)
+
+// VerificationResult contains the result of a verification of a certificate
+// against a browser. It discerns between NameError, meaning the name on the
+// certificate does not match, and a ValidationError, meaning some sort of issue
```
zakird: A certificate being expired is really none of these. How would that fit in here? What about key usage problems?

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// True is the certificate is blacklisted by a given browser as untrusted,
+	// e.g. Cloudflare certificates valid at the time of Heartbleed disclosure in
+	// Chrome.
+	Blacklisted bool
+
+	// ValiditionError will be non-nil when there was some sort of error during
+	// validation not involving a name mismatch, e.g. if a chain could not be
+	// built.
+	ValidationError error
```
zakird: This should be an enum not a string.

x509/cert_pool.go
```diff
@@ -128,3 +128,7 @@ func (s *CertPool) Subjects() (res [][]byte) {
 	}
 	return
 }
+
+func (s *CertPool) Contains(c *Certificate) bool {
```
zakird: This is pretty poor form to inject into an independent library. If this isn't yet implemented, this needs to panic, or we shouldn't add it. But as it stands, it's pretty dangerous and unneeded to add this here.

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+// certificate in Intermediates, and returns all such chains. It additional
+// checks if the Name in the VerificationOptions matches the name on the
+// certificate. Finally, it checks to see if c is blacklisted or whitelisted by
+// the Browser.
+func (b *Browser) Verify(c *x509.Certificate, opts VerificationOptions) (res *VerificationResult) {
+	res = new(VerificationResult)
+	res.Name = opts.Name
+
+	// Always run prevalidate.
+	if err := v.VerifyProcedure.PreValidate(c); err != nil {
```
zakird: What do you envision going in pre-validate? Versus Post-validate? Why is that relevant if the point of this function is to see if it links up to the root, not to validate the certificate?

verifier/verifier.go
```diff
@@ -0,0 +1,142 @@
...
+		res.ValidationError = err
+	}
+
+	// XXX: x509 should expose the validation we want with out this struct
+	// designed for JSON output.
+	if res.ValidationError == nil {
+		xopts := v.convertOptions(&opts)
+		chains, validation, _ := c.ValidateWithStupidDetail(xopts)
```
zakird: This function name is non-descriptive and unhelpful.

zakird:
I left a bunch of comments, but I'm very much not convinced this is the right form of doing any of this. There seems to be a mismatch between how much detail goes in and out, and why pieces are done in any location. I think it might make sense to write this out once at a higher level before implementing the scaffolding.

verifier/verifier.go
zakird: Why doesn't this take a chain?

verifier/verifier.go
dadrian: Why wouldn't you return this?

verifier/verifier.go
dadrian: Then both bools are true?

verifier/verifier.go
dadrian: There may be multiple chains to the same root store through different intermediates, and you may want to apply different validation (e.g. remove any chain that contains SHA1).

verifier/verifier.go
dadrian: Chains haven't been built yet, but you raise a good point that "served chains" aren't easily passed in here.

zakird:
I'm not sure why you'd pass out something you just passed in.

verifier/verifier.go
dadrian: This is just compatability for calling down to the actual x509 lib.

verifier/verifier.go
dadrian: Which is why I left it simple for now? We can extract more interfaces later.

verifier/verifier.go
dadrian: This is the top level function, I don't understand the problem here? At some point, you need to get all this information back per-store.

verifier/verifier.go
dadrian: This is not a string, it's an error object.

verifier/verifier.go
dadrian: That function existed already, note the TODO.

verifier/verifier.go
dadrian: It seemed reasonable to add a hook.

x509/cert_pool.go
dadrian: This is still just framing, I'm not suggesting we merge this.

verifier/verifier.go
dadrian: Agreed. We should handle expiration separately. A key usage problem would be a ValidationError.

dadrian:
I think the confusion here is that I am considering this to be a high-level interface, where a root store is a root store, a set of intermediates likely trusted by that root store, possible additional validation logic (e.g. `PreVerify` and `PostVerify`), a mechanism for whitelisting, and a mechanism for blacklisting. The goal here was to collect all of the information needed to populate the `RootStoreStatus` struct in https://github.com/research/censys-definitions, except expose it using native/idiomatic Golang code and types.

The Verifier object is meant to handle the interactions with the x509 library, take certificates in and provide validation data out, and allow the necessary hooks to implement anything that's custom per-root store / browser / operating system being mimicked. This is the top-level interface for implementing multi-validation, but would not be what a user of this library would interact with directly (see comments below about pre-defined `Verifier` objects).

Now, there's obviously still some issues here, some I left intentionally, and others that you brought up.

- Many aspects are still left as exercises for the reader. This is intentional. I was just trying to lay out some example logic that was capable of doing the necessary checks we want for verification, and exposing all the data in the end result.
    - Creating a root and intermediate store is still an exercise left for the reader.
    - Implementing `PreVerify` and `PostVerify` is also an exercise left for the reader. A completed library would provide implementations of these as needed, and have pre-defined `Verifier` objects corresponding with different browsers and root stores.
- Blacklist and Whitelist is just a call to `Contains`. This seemed like an easy way to show how blacklisting and whitelisting fit into the logic flow, without having to dig in and define some abstract  interface. At some point, we'll probably want to abstract it out more, but as mentioned, it doesn't seem very necessary right now, and this seemed like a decent stub.
- I didn't handle expiration well, or provide a way to pass in served chains. This was a mistake, and why I submitted this for comments already. 😄

PR #18, opened 2017-03-20
Add NeverValid Certificate InvalidReason

dadrian:
This unifies all expiration handling into one check that allows
certificates with chains that were at some point valid (e.g. expired),
to be discerned from certificates with chains that were never possibly
valid (e.g. the dates did not overlap, but the chain was otherwise
valid).

This will result in longer calls to Verify(), since we disqualify some
(many?) chains later.

Additionally, this ensures that hostname verification is always called
last during validation. This makes it easier to discern HostnameErrors
from other errors, although it may result in longer calls to Verify().

This adds two test certificates for NeverValid, and a script to generate
them in x509/testdata/gen_zcrypto.sh.


dadrian:
DO NOT MERGE

dadrian:
This adds a new `InvalidReason` used with `CertificateInvalidError` for `NeverValid`, which represents certificates that could not possibly have been valid at any point in time for date-related issues. It modifies the Verification process so that date-errors are checked last.

I added two tests for `NeverValid` certificates: one that has a leaf certificate with `NotBefore > NotAfter`, and another that where the leaf is only valid outside of the intermediates validity window.

Some questions that may still need answering: 
 - How does this interact with separating out Hostname errors from other errors?
 - What happens if there are two chains, one of which is NeverValid and another which is CurrentlyValid / WasValid. To be honest, I'm not sure how to structure the chains to write this test (in terms of Intermediate subjects / keys / leaf AKID's, etc.).

zakird:
This looks fine to me.

dadrian:
@zakird I think this is ready to go now

PR #21, opened 2017-04-06
Remove NameConstraint checking from x509.Verify()

dadrian:
This updates x509.Verify() to no longer enforce name constraints when
passed an opts.DNSName. This was a haphazard at best implementation, and
did not fully enforce the connection anyway. NameConstraints are likely
best checked after chains have been built, and can be dependent on the
name being verified. They might also be enforced differently in
different stores. Therefore, it's probably best not to check them here
until we decide on how we want to enforce them.


x509/extensions.go
```diff
@@ -120,6 +120,19 @@ func (cp *CertificatePoliciesData) MarshalJSON() ([]byte, error) {
 	return json.Marshal(policies)
 }
 
+// GeneralNames corresponds an X.509 GeneralName defined in
+// Section 4.2.1.6 of RFC 5280.
+//
+// GeneralName ::= CHOICE {
```
zakird: Not sure why these are commented out?

dadrian: That's not Go code, it's a comment with documentation copy/pasted from the RFC? That's ASN1 notation :)

PR #22, opened 2017-04-11
Add initial zcrypto/verifier

dadrian:
Verifier is a higher-level module for doing in-depth certificate
verification, with the goal of being able to impersonate a browser or
specific root store's validation procedure. It uses the chaining code in
x509 to generate chains, and then apply's customizable verification on
top.

This includes a few "built-in" verifiers meant to represent the Firefox,
Microsoft, Apple, and Java trust stores.

The package is not yet complete, but should have roughly equivalent
functionality to x509.ValidateWithStupidDetail, which is now deprecated.

### DIFF
--- a/verifier/builtins.go
+++ b/verifier/builtins.go
@@ -1,0 +1,60 @@
+/*
+ * ZCrypto Copyright 2017 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package verifier
+
+import "github.com/zmap/zcrypto/x509"
+
+// Built-in verifiers representing common root stores.
+var (
+	// NSS is a Verifier mimicking the validation used in Firefox.
+	NSS Verifier
+
+	// Microsoft is a Verifier mimicing the validation in Windows 10 SChannel.
+	Microsoft Verifier
+
+	// Apple is a Verifier mimicing the validation in OS X Sierra SecureTransport.
+	Apple Verifier
+
+	// Java is a Verifier mimicing the validation in Java 8 javax.net.ssl.
+	Java Verifier
+)
+
+// InitializeNSS sets up the built-in NSS Verifier.
+func InitializeNSS(roots, intermediates *x509.CertPool) {
+	NSS.Roots = roots
+	NSS.Intermediates = intermediates
+	NSS.VerifyProcedure = &VerifyProcedureNSS{}
+}
+
+// InitializeMicrosoft sets up the built-in Microsoft Verifier.
+func InitializeMicrosoft(roots, intermediates *x509.CertPool) {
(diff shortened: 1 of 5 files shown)


verifier/verifier.go
```diff
@@ -0,0 +1,133 @@
...
+	out.Intermediates = v.Intermediates
+	out.DNSName = opt.Name
+	return
+}
+
+// Verify checks if c chains back to a certificate in Roots, possibly via a
+// certificate in Intermediates, and returns all such chains. It additional
+// checks if the Name in the VerificationOptions matches the name on the
+// certificate. Finally, it checks to see if c is blacklisted or whitelisted by
+// the Verifier.
```
bvandersloot: Do you want to check BasicConstraints as well here?

verifier/verifier.go
```diff
@@ -0,0 +1,133 @@
...
+// chain.
+type VerificationResult struct {
+
+	// The name being checked, e.g. "www.censys.io".
+	Name string
+
+	// Whitelisted is true if the certificate is whitelisted by a given verifier
+	// as trusted, e.g. when Apple whitelisted subset of StartCom certs by SPKI
+	// hash in the wake of the WoSign incidents surrounding misissuance.
+	Whitelisted bool
```
bvandersloot: Maybe a slice of errors, rather than 3 independent booleans and two errors? I haven't seen that pattern before, but I also dislike looking in N places for a validation result

verifier/builtins.go
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package verifier
+
+import "github.com/zmap/zcrypto/x509"
+
+// Built-in verifiers representing common root stores.
+var (
+	// NSS is a Verifier mimicing the validation used in Firefox.
+	NSS Verifier
```
bvandersloot: Maybe put these in a map that is keyed by an enum? I'm not set on it, but if you wanted to do all of these at once, it would make it a bit easier

verifier/builtins.go
```diff
@@ -0,0 +1,60 @@
...
+ * permissions and limitations under the License.
+ */
+
+package verifier
+
+import "github.com/zmap/zcrypto/x509"
+
+// Built-in verifiers representing common root stores.
+var (
+	// NSS is a Verifier mimicing the validation used in Firefox.
```
ajholland: https://img.memesuper.com/77a96c93ce547992cb7f93ffb2f9ac65_you-had-a-spelling-error-but-you-made-me-laugh-approving-spelling-error-meme_600-600.jpeg

verifier/verifier.go
```diff
@@ -0,0 +1,133 @@
...
+
+// VerifyProcedure is an interface to implement additional browser specific logic at
+// the start and end of verification.
+type VerifyProcedure interface {
+	// TODO
+}
+
+// VerificationOptions contains settings for Verifier.Verify().
+// VerificationOptions should be safely copyable.
+type VerificationOptions struct {
```
bvandersloot: Potentially, this struct could use KeyUsages. Not sure if we would ever need to do this, but you may want to include it for completeness sake

verifier/verifier.go
dadrian: I believe that's checked in `x509.Verify`. I should remove the comment about blacklisted vs whitelisted though.

verifier/verifier.go
dadrian: FUTURE WORK

verifier/verifier.go
dadrian: Yeah, I agree that's a problem, but this is meant to mimic data that's going to get fed into Censys. We should probably add a meta-result ("would this load without an error"), but I'm not sure how to expose it yet.

PR #24, opened 2017-04-13
zintermediate wip

dadrian:
no merge


dadrian:
This has an infinite loop. Debugging is required.

PR #25, opened 2017-04-14
Add SubjectEquals to x509.Certificate

dadrian:
Avoid duplicates in chains by using SubjectEquals in Verify, instead of
Equals.


zakird:
This code looks fine. However, can you explain why this is the right change to be making?

zakird:
@dadrian can you explain or close?

dadrian:
This prevents infinite loops in chain building in the event you add several intermediates with the same subject to your intermediate cert pool, with a cross-signing loop somewhere.

PR #27, opened 2017-04-18
Kumarde/ct client

kumarde:
Adding a simple consumer of the ct client that grabs a *x509.Certificate from a log given an Index.


ct/client/main/main.go
```diff
@@ -8,17 +8,14 @@ import (
 	"encoding/json"
 	ct "github.com/zmap/zcrypto/ct"
 	"github.com/zmap/zcrypto/ct/client"
-	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zcrypto/ct/x509"
```
zakird: Are you sure this is what you want?

kumarde: I did this because it's currently what ctsync pull_daemon uses to parse certificates. Since I was using this to figure out what pull_daemon was missing, this is why I modified it.

zakird: It's used, but then it's not used, and the other one is used.

kumarde: Oh, I see. Because the handler.go file uses zcrypto/x509. GAH

zakird: welcome to ctsync.

kumarde: Well, disregard everything I said about the stuff I ran last night then :)

ct/client/main/main.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"encoding/json"
+	ct "github.com/zmap/zcrypto/ct"
+	"github.com/zmap/zcrypto/ct/client"
+	"github.com/zmap/zcrypto/x509"
+)
+
+const (
+	// matchesNothingRegex is a regex which cannot match any input.
+	matchesNothingRegex = "a^"
```
zakird: Huh? This doesn't look used anywhere.

kumarde:
@zakird whoops, you're right.

PR #28, opened 2017-04-18
adding TBS parsing function

kumarde:
This is a ct/x509.go function that is used exclusively for parsing Precerts, that we will need to grab the certs we're currently missing from CT.


zakird:
I don't think is is the right approach. Instead we should just fix our parse certificate call to handle precerts.

zakird:
I'm closing this since we decided this is not the correct approach off channel.

PR #34, opened 2017-04-26
Reland 08aab958ab1686d45518792a285e191c664a16df

dadrian:
This was not the root of the validation bugs, we had known-good scans
since this had originally landed.

-----

Add NeverValid Certificate InvalidReason (#18)

This unifies all expiration handling into one check that allows
certificates with chains that were at some point valid (e.g. expired),
to be discerned from certificates with chains that were never possibly
valid (e.g. the dates did not overlap, but the chain was otherwise
valid).

This will result in longer calls to Verify(), since we disqualify some
(many?) chains later.

Additionally, this ensures that hostname verification is always called
last during validation. This makes it easier to discern HostnameErrors
from other errors, although it may result in longer calls to Verify().

This adds two test certificates for NeverValid, and a script to generate
them in x509/testdata/gen_zcrypto.sh.


zakird:
sgtm

PR #35, opened 2017-04-26
Reland fa7c8670d3aa365afff30c80c3d5ed39331e9670

dadrian:
This commit didn't directly touch x509.ValidateWithStupidDetail, and
there only affected code paths in Verify() where opts.DNSName or
opts.EmailAddress were set. Tests pass, and ZGrab behavior is
reasonable.

--------

Remove NameConstraint checking from x509.Verify() (#21)

This updates x509.Verify() to no longer enforce name constraints when
passed an opts.DNSName. This was a haphazard at best implementation, and
did not fully enforce the connection anyway. NameConstraints are likely
best checked after chains have been built, and can be dependent on the
name being verified. They might also be enforced differently in
different stores. Therefore, it's probably best not to check them here
until we decide on how we want to enforce them.


zakird:
sgtm

PR #38, opened 2017-04-26
Set BrowserError when validation fails

dadrian:
We also were only setting browser error when matches_domain was checked and also false. Failing to set browser error masked the bug in ZTag


dadrian:
Rather, setting browser_error masked the bug in ZTag

PR #40, opened 2017-05-03
Limit max number of intermediates

dadrian:
This cleans up the verification function, prevents roots that are also
intermediates from appearing in a chain twice, and limits the max number
of intermediates to 10. This makes zcrypto/x509 suitable for
intermediate trawling with Censys.

### DIFF
--- a/x509/cert_pool.go
+++ b/x509/cert_pool.go
@@ -119,6 +119,20 @@ func (s *CertPool) AppendCertsFromPEM(pemCerts []byte) (ok bool) {
 	return
 }
 
+// Contains returns true if c is in s.
+func (s *CertPool) Contains(c *Certificate) bool {
+	candidates, ok := s.byName[string(c.RawSubject)]
+	if !ok {
+		return false
+	}
+	for _, candidateNum := range candidates {
+		if s.certs[candidateNum].Equal(c) {
+			return true
+		}
+	}
+	return false
+}
+
 // Subjects returns a list of the DER-encoded subjects of
 // all of the certificates in the pool.
 func (s *CertPool) Subjects() (res [][]byte) {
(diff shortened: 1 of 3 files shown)


x509/cert_pool.go
```diff
@@ -119,6 +119,20 @@ func (s *CertPool) AppendCertsFromPEM(pemCerts []byte) (ok bool) {
 	return
 }
 
+// Contains returns true if c is in s.
+func (s *CertPool) Contains(c *Certificate) bool {
+	candidates, ok := s.byName[string(c.RawSubject)]
+	if !ok {
+		return false
+	}
+	for _, candidateNum := range candidates {
```
zakird: do you end up with a copy if you do this the other way, or why are you doing this this way?

x509/verify.go
```diff
@@ -60,6 +60,10 @@ func (c *Certificate) isValid(certType CertificateType, currentChain []*Certific
 		}
 	}
 
+	if len(currentChain) > 10 {
```
zakird: this feels like it should be a constant.

x509/cert_pool.go
```diff
@@ -128,3 +142,10 @@ func (s *CertPool) Subjects() (res [][]byte) {
 	}
 	return
 }
+
+// Certificates returns a list of parsed certificates in the pool.
+func (s *CertPool) Certificates() []*Certificate {
+	out := make([]*Certificate, 0, len(s.certs))
```
zakird: they're pointers anyway. why not just return s.certs?

x509/cert_pool.go
dadrian: I don't want indicies of this array, I want the values of this array, which happen to be indicies for a different array.

x509/cert_pool.go
dadrian: Because you can still break up the underlying slice in the CertPool since slices are quasi-by-reference.

x509/verify.go
dadrian: You feel like a constant.

Fine...

dadrian:
Good day, sir!

zakird:
and good riddance to you!

dadrian:
Righto!

PR #42, opened 2017-05-03
Add zintermediate utility

dadrian:
This adds an executable that takes in a large number (e.g. 120K) of
possible intermediates, and checks to see if they're actually
intermediates.

### DIFF
--- a/x509/zintermediate/zintermediate.go
+++ b/x509/zintermediate/zintermediate.go
@@ -1,0 +1,185 @@
+/*
+ * ZCrypto Copyright 2017 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package main
+
+import (
+	"bufio"
+	"encoding/base64"
+	"encoding/pem"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"time"
+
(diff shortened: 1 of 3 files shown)


verifier/verifier.go
```diff
@@ -104,6 +104,7 @@ func (v *Verifier) convertOptions(opt *VerificationOptions) (out x509.VerifyOpti
 	out.Roots = v.Roots
 	out.Intermediates = v.Intermediates
 	out.DNSName = opt.Name
+	out.KeyUsages = []x509.ExtKeyUsage{x509.ExtKeyUsageAny}
```
zakird: feels like this should be called ExtendedKeyUsages given that KeyUsages exists and is something else.

dadrian: This matches x509.VerifyOptions, and tbh, should probably be in a different commit.

x509/zintermediate/zintermediate.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	inputFileName = flag.Arg(0)
+
+	inputFormatArg = strings.ToLower(inputFormatArg)
+	switch inputFormatArg {
+	case "base64":
+		inputFormat = inputFormatBase64
+	case "pem":
+		inputFormat = inputFormatPEM
+	case "json":
```
zakird: oh gawd why

dadrian: Because you get certs from Censys. It would just read `raw`, if I got around to it.

verifier/verifier.go
zakird: What calls this and why is this set here? It's not called by the other code?

x509/zintermediate/zintermediate.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		log.Fatalf("could not open %s: %s", inputFileName, err)
+	}
+	rootPool, err := loadPEMPool(rootFile)
+	rootFile.Close()
+	if err != nil {
+		log.Fatalf("could not load roots: %s", err)
+	}
+
+	log.Infof("loading candidate intermediates from %s", inputFileName)
```
zakird: This should be on line 122 given the location of L117.

verifier/verifier.go
dadrian: Nothing calls this right now. I'm going to remove this change from this commit.

PR #43, opened 2017-05-03
Check self-signed during parsing

dadrian:
Correctly detect self-signed certificates by comparing raw subjects to
raw issuers, and verifying the signature on the certificate against
itself. This happens automatically during parsing, and is exposed as a
bool.


dadrian:
This even has tests! That I happened to write while employed by Google!

PR #44, opened 2017-05-04
Add a validation error for non-root self-signed certificates

dadrian:
This explicits marks root certificates as having valid chains of length
1, and marks other non-root self-signed certificates as
CertificateInvalidError with the reason IsSelfSigned. This saves us a
few crypto operations, but more importantly gives us the correct
validation result for root certificates.

### DIFF
--- a/x509/verify.go
+++ b/x509/verify.go
@@ -173,6 +173,17 @@ func (c *Certificate) inChain(chain []*Certificate) bool {
 
 func (c *Certificate) buildChains(cache map[int][][]*Certificate, currentChain []*Certificate, opts *VerifyOptions) (chains [][]*Certificate, err error) {
 
+	// If the certificate being validated is a root, add the chain of length one
+	// containing just the root. Only do this on the first call to buildChains,
+	// when the len(currentChain) = 1.
+	if len(currentChain) == 1 && opts.Roots.Contains(c) {
+		chains = append(chains, appendToFreshChain(nil, c))
+	}
+
+	if len(chains) == 0 && c.SelfSigned {
+		err = CertificateInvalidError{c, IsSelfSigned}
+	}
+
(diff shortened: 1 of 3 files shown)


x509/verify.go
```diff
@@ -173,6 +173,14 @@ func (c *Certificate) inChain(chain []*Certificate) bool {
 
 func (c *Certificate) buildChains(cache map[int][][]*Certificate, currentChain []*Certificate, opts *VerifyOptions) (chains [][]*Certificate, err error) {
 
+	if len(currentChain) == 1 && opts.Roots.Contains(c) {
```
zakird: I don't quite understand the logic here. Can you add in a comment?

dadrian: Done.

x509/verify.go
```diff
@@ -173,6 +173,17 @@ func (c *Certificate) inChain(chain []*Certificate) bool {
 
 func (c *Certificate) buildChains(cache map[int][][]*Certificate, currentChain []*Certificate, opts *VerifyOptions) (chains [][]*Certificate, err error) {
 
+	// If the certificate being validated is a root, add the chain of length one
+	// containing just the root. Only do this on the first call to buildChains,
+	// when the len(currentChain) = 1.
```
zakird: But _why_ are you doing this?

dadrian: Which part of the _why_ is confusing? Is it why do return a chain of length 1, or why isn't this somehow handled later in the function?

zakird:
This code seems fine, but it would help to expand a bit on the comment. It's not immediately clear why this change is necessary.

x509/verify.go
zakird: Why are you adding the root into the chain?

PR #47, opened 2017-05-08
Expand VerifyResult

dadrian:
- Add information about expiration (InValidityWindow)
- Add HasValidChain() and HadValidChain() methods
- Add certificate type

### DIFF
--- a/verifier/verifier.go
+++ b/verifier/verifier.go
@@ -57,17 +57,17 @@ type VerificationResult struct {
 	// of the certificate being verified.
 	Parents []*x509.Certificate
 
-	// Current is a list of validated certificate chains that are valid at
+	// CurrentChains is a list of validated certificate chains that are valid at
 	// ValidationTime, starting at the certificate being verified, and ending at a
 	// certificate in the root store.
 	CurrentChains [][]*x509.Certificate
 
-	// Expired is a list of certificate chains that were valid at some point,
-	// but not at ValidationTime.
+	// ExpiredChains is a list of certificate chains that were valid at some
+	// point, but not at ValidationTime.
 	ExpiredChains [][]*x509.Certificate
 
-	// Never is a list of certificate chains that could never be valid due to
-	// date-related issues, but are otherwise valid.
+	// NeverValidChains is a list of certificate chains that could never be valid
+	// due to date-related issues, but are otherwise valid.
 	NeverValidChains [][]*x509.Certificate
 
 	// CertificateType is one of Leaf, Intermediate, or Root.
--- a/verifier/verifier.go
+++ b/verifier/verifier.go
@@ -53,18 +53,22 @@ type VerificationResult struct {
 	// certificate and the name being verified against.
 	NameError error
 
+	// Parents is a set of currently valid certificates that are immediate parents
+	// of the certificate being verified.
+	Parents []*x509.Certificate
+
 	// Current is a list of validated certificate chains that are valid at
 	// ValidationTime, starting at the certificate being verified, and ending at a
 	// certificate in the root store.
-	Current [][]*x509.Certificate
+	CurrentChains [][]*x509.Certificate
 
 	// Expired is a list of certificate chains that were valid at some point,
 	// but not at ValidationTime.
-	Expired [][]*x509.Certificate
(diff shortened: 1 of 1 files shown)


dadrian:
@zakird we need this for the cert daemon, but we don't have tests for this high-level package. I could spend time converting all the tests in the x509 package to be able to be used by both this high-level verifier package, and the x509.Verify() function, or we could just go for it.

verifier/verifier.go
```diff
@@ -65,13 +65,35 @@ type VerificationResult struct {
 	// Never is a list of certificate chains that could never be valid due to
 	// date-related issues, but are otherwise valid.
 	Never [][]*x509.Certificate
+
+	// CertificateType is one of Leaf, Intermediate, or Root.
+	CertificateType x509.CertificateType
+
+	// VerifyTime is time used in verification, set in the VerificationOptions.
+	VerifyTime time.Time
```
zakird: I don't quite understand why we pass this back but _shrug_

verifier/verifier.go
```diff
@@ -65,13 +65,35 @@ type VerificationResult struct {
...
+
+	// InValidityWindow is true if NotBefore < VerifyTime < NotAfter
+	InValidityWindow bool
```
zakird: Why don't we just call this expired?

verifier/verifier.go
```diff
@@ -65,13 +65,35 @@ type VerificationResult struct {
...
+func (res *VerificationResult) HasTrustedChain() bool {
+	return len(res.Current) > 0
+}
+
+// HadTrustedChain returns true if at some point in time, the certificate had a
+// chain to a trusted root in this store.
+//
+// This is equivalent to checking if len(current) > 0 || len(expired) > 0
+func (res *VerificationResult) HadTrustedChain() bool {
+	return res.HasTrustedChain() || len(res.Expired) > 0
```
zakird: I would have expected Expired to be a boolean that described whether the certificate was expired. Perhaps we should find a more descriptive term for this field.

verifier/verifier.go
```diff
@@ -113,20 +134,35 @@ func (v *Verifier) convertOptions(opt *VerificationOptions) (out x509.VerifyOpti
...
 
-		res.Current, res.Expired, res.Never, res.ValidationError = c.Verify(xopts)
-		if len(opts.Name) > 0 {
-			res.NameError = c.VerifyHostname(opts.Name)
-		}
+	// Determine certificate type.
+	if xopts.Roots.Contains(c) {
+		// A certificate is only a root if it's in the root store.
+		res.CertificateType = x509.CertificateTypeRoot
+	} else if c.IsCA && !c.SelfSigned {
```
zakird: You could be self signed and also be an intermediate.

dadrian: How?

zakird: C1 has K1 , subj=S1, issuer=I1, S1==I1, sn=SN1, is a root
C2 has K1, S1, I1, sn=SN2, is not a root. 
C3 is signed by C2. Could not have been signed by C1 because AKID includes C2 SN.

dadrian: What is SN and how is it "included in" AKID? Are you suggesting a chain of C3 -> C2 -> C1?

zakird: Yes

dadrian: What's the AKID in C2?

zakird: Not entirely sure how to answer your question about how it's included in the AKID field. You might want to read Section 4.2.1.1 of RFC 3280.

zakird: Let's say AKID is null in C2.

dadrian: That RFC suggests content of AKID is largely "whatever you want derived from public key", but we can imagine a situation where this is the case. I think our chaining is built entirely on the keyIdentifier sequence element anyway. You don't happen to know of a chain that does this, do you?

As for setting the certificate type, should we just set anything that's not a root with IsCA = true as an intermediate? This will include everybody who just ran `openssl req -new -x509`.

zakird: I don't know of one offhand, but I know at least one exists, because we've hit it before.

PR #48, opened 2017-05-10
Fix external ClientHello import

mtury:
Hi,
This is a fix for https://github.com/zmap/zgrab/pull/223
The connection config needs to be updated after a raw ClientHello import, as when using a ClientHello fingerprint. Unless we do that, there is a high chance of discrepancies between the ServerHello received (based on the ClientHello we sent) and whatever zgrab is ready to negotiate (based on the config)!
The patch seems pretty straightforward but I haven't checked every field with the right test connections yet. At least I could see it worked as expected when providing cipher suites different from the defaultConfig ones. ;)

### DIFF
--- a/tls/handshake_client.go
+++ b/tls/handshake_client.go
@@ -84,6 +84,20 @@ func (c *ClientFingerprintConfiguration) CheckImplementedExtensions() error {
 	return nil
 }
 
+func (c *clientHelloMsg) WriteToConfig(config *Config) error {
+	config.NextProtos = c.alpnProtocols
+	config.CipherSuites = c.cipherSuites
+	config.MaxVersion = c.vers
+	config.ClientRandom = c.random
+	config.CurvePreferences = c.supportedCurves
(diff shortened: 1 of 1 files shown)


dadrian:
Offhand the patch seems fine. @bvandersloot, can you review since you wrote the original `ClientFingerprintConfiguration` code?

@mtury it looks like this caused a test to fail, I'm not sure why. Can you investigate?

bvandersloot:
I don't believe this touches any of my codepaths or fails my tests. It just looks related to `ClientFingerprintConfiguration` because of where the first function was inserted and how GitHub rendered this.

mtury:
As @bvandersloot says, the patch sits next to the fingerprint code (and it looks similar), but actually their execution is mutually exclusive.

There is no failing test when I run it on my side, and one of the fails here seems to be related to a time sensitive TCP connection. Could you restart the build maybe?

dadrian:
Thanks!

mtury:
Well thank you for your work here. :)

PR #50, opened 2017-05-11
Ensure ValidateWithStupidDetail returns all errors

dadrian:
Previously, if chain building failed but hostname verification
succeeded, ValideWithStupidDetail would incorrectly return a nil error
object, despite returning no chains and correctly populating the
Validation object.


dadrian:
@SergeyFrolov I think this fixes your problem in the Decoy Routing client.

PR #54, opened 2017-05-15
Pass presented chain to Verifiers

dadrian:
This approach is thread safe and allows the intermediate set used by a
Verifier to be appended with the chain presented by the host.

### DIFF
--- a/verifier/verifier.go
+++ b/verifier/verifier.go
@@ -107,8 +107,9 @@ type VerifyProcedure interface {
 // VerificationOptions contains settings for Verifier.Verify().
 // VerificationOptions should be safely copyable.
 type VerificationOptions struct {
-	VerifyTime time.Time
-	Name       string
+	VerifyTime     time.Time
+	Name           string
+	PresentedChain *x509.CertPool
 }
 
 func (opt *VerificationOptions) clean() {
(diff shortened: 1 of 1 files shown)


dadrian:
This depends on #53.

verifier/verifier.go
```diff
@@ -127,7 +128,11 @@ type Verifier struct {
 func (v *Verifier) convertOptions(opt *VerificationOptions) (out x509.VerifyOptions) {
 	out.CurrentTime = opt.VerifyTime
 	out.Roots = v.Roots
-	out.Intermediates = v.Intermediates
+	if opt.PresentedChain != nil && opt.PresentedChain.Size() > 0 {
+		out.Intermediates = v.Intermediates.Sum(opt.PresentedChain)
```
zakird: It's not clear to me that this is the correct approach. What performance impact does this have? Is this better or worse than adding and removing the chain from the cert pool?

dadrian: We can't add and remove the chain from the cert pool while being thread safe in the current setup.

zakird: Alright. What's the performance impact?

dadrian: Remove logic would have to reindex everything "after" the certificate being removed in two maps. In the common case this would be quick (remove the last certificate), but it's still pretty gnarly. There's a bunch of edge cases in the removal code (what if multiple certs have the same subject and you're removing only one of them) and in it's usage (what if the presented chain certificate was already in the pool and we remove it?)

dadrian: We have no idea what the performance impact of this change is, in the same way we have no idea what the performance of this code before the change is. If we're too slow, we can profile. This feels like optimizing too early.

PR #59, opened 2017-05-24
Match pkix.Name.String() ordering to ToRDNSequence()

dadrian:


dadrian:
Nope! We don't use that to print anything out.

PR #61, opened 2017-05-26
Deduplicate in chain building by (name, spki)

dadrian:
Previously we deduplicated by full certificate

### DIFF
--- a/x509/verify.go
+++ b/x509/verify.go
@@ -209,7 +209,7 @@ func (c *Certificate) buildChains(cache map[int][][]*Certificate, currentChain [
 		if err != nil {
 			continue
 		}
-		if !root.subjectAndSPKIInChain(currentChain) {
+		if !root.inChain(currentChain) {
 			chains = append(chains, appendToFreshChain(currentChain, root))
 		}
 	}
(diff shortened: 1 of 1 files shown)


x509/verify.go
```diff
@@ -198,7 +209,7 @@ func (c *Certificate) buildChains(cache map[int][][]*Certificate, currentChain [
 		if err != nil {
 			continue
 		}
-		if !root.inChain(currentChain) {
+		if !root.subjectAndSPKIInChain(currentChain) {
```
dadrian: I'm wondering if this should remain `inChain` for roots.

zakird: I'm not sure. I didn't merge this based on this comment, but I'm not sure if this is blocking or not. Can you resolve one way or another?

dadrian: Thinking about this a bit more, I think `inChain` is probably correct. Consider a case where there's a "newer" version of a root that exists as both an intermediate with a parent of the old root, and a self-signed "new root". We'd never chain through the intermediate back to the old root.

Now, this might actually be invalid / not something that happens.

Issue #63, opened 2017-05-31
Certificate.CheckSignatureFrom should compare bytes


dadrian:
https://github.com/zmap/zcrypto/blob/9791508aa1693598f7732f32e66503d3a70fc969/ct/x509/x509.go#L622

This should use `bytes.Equal`, see https://github.com/golang/go/blob/master/src/crypto/x509/verify.go#L194-L199 in Golang upstream.

Then again, maybe we don't actually want to do this at all. Windows doesn't!

dadrian:
Fixed in #67 / cce896947854c0fe8bdac183860948c1634c8633

PR #65, opened 2017-06-02
Remove buggy subject/issuer comparison

dadrian:
This is one approach to resolving #63. We could switch it to `bytes.Equal(c.RawIssuer, parent.RawSubject)`, but given that not all crypto libraries do this, I suggest we start with a broad approach and then narrow it down to match specific stores. E.g. for "Firefox" or "Chrome", additionally require that issuer byte-for-byte match subject. But for Microsoft, don't require this.

Note that when building chains we do require subjects to match if there is no AKID/SKID. This "don't check if they match" commit only changes behavior for certificates that match on AKID/SKID, but don't match on Subject/Issuer.

### DIFF
--- a/x509/x509.go
+++ b/x509/x509.go
@@ -751,10 +751,6 @@ func (c *Certificate) CheckSignatureFrom(parent *Certificate) (err error) {
 
 	// TODO(agl): don't ignore the path length constraint.
 
-	if parent.Subject.String() != c.Issuer.String() {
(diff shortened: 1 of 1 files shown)


zakird:
I disagree with this unless we can get the extra check added to NSS at the same time. I think it's very confusing and we're introducing further incorrect behavior for the root store we currently have deployed.

PR #68, opened 2017-06-06
Calculate parents at time of expiration

dadrian:


verifier/verifier.go
```diff
@@ -95,7 +95,7 @@ func (res *VerificationResult) HasTrustedChain() bool {
 //
 // This is equivalent to checking if len(current) > 0 || len(expired) > 0
 func (res *VerificationResult) HadTrustedChain() bool {
-	return res.HasTrustedChain() || len(res.ExpiredChains) > 0
+	return res.HasTrustedChain() || len(res.ExpiredChains) > 0 || len(res.Parents) > 0
```
dadrian: This part is a little more bold. Thoughts?

PR #82, opened 2017-07-17
Match pkix.Name.String() order to certificate

### LINKED ISSUE #81: subject_dn should not use static order
Right now, the `parsed.subject_dn` and `parsed.issuer_dn` fields are output in a static order. They should be output as the fields are ordered in the certificate.

dadrian:
When possible, just use Name.OriginalRDNSequence instead of rebuilding
the whole sequence from scratch. This way we support creating new names
and turning them into sequences, as well as reading in a Name, and
outputing it back identically.

Fixes #81

### DIFF
--- a/x509/extensions_test.go
+++ b/x509/extensions_test.go
@@ -5,144 +5,17 @@
 package x509
 
 import (
-	"encoding/json"
-	"encoding/pem"
-	"fmt"
-	"io/ioutil"
-	"strings"
 	"testing"
-
-	. "gopkg.in/check.v1"
 )
 
-func TestExtensions(t *testing.T) { TestingT(t) }
-
-type ExtensionsSuite struct {
-	pemData     []byte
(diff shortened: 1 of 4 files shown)


zakird:
seems fine minus tests not passing

dadrian:
Rather than attempt to fix all the tests that relied on obscure implementation details to pass, I removed the poorly designed tests and left detailed TODOs. I will file relevant issues in this repository of things to fix. Largely, the failing tests were indicative of code that needs to be refactored (both in the tests and in the base library), and not of any bugs.

Rather than extend this PR even further, I'm going to document the issues for later and move on.

dadrian:
@cdzombak I'd appreciate a second pass just to make sure I'm not crazy.

dadrian:
Filed https://github.com/zmap/zcrypto/issues/83

PR #85, opened 2017-07-31
Add Graph model of PKI used by Verifier package

dadrian:
This is a behemoth of a PR that introduces `x509.Graph` and associated types. A `Graph` is a math-style multigraph where nodes are `(key, subject)` tuples, and edges are certificates. The head/source of an edge is the issuer, and the tail/destination is the subject.

A Graph can be walked to find chains from a certificate to a root. This certificate may or may not already be in the graph. Graphs do not yet support a walk given a second, possibly overlapping, `PresentedChain` graph.

This updates the ZCrypto-introduced `verifier` package to use the Graph-model for chain building. It leaves the golang-fork x509.Verify function as is. Longer term, once we support PresentedChain, the plan would be to remove the original Golang Verify function and only have the ZCrypto `verifier`. It also consolidates types within the x509 package as needed (e.g. introduces `CertificateChain`).

`verifier` still has no tests, although the graph walk itself is tested.

I'm going to go through this myself and inevitably find silly things or things that need more documentation or clean up, and will be tackling those, but in the mean time, I'd like to get the ball rolling on reviews from other people, since certificate validation is hard and graphs are complicated and everything is on fire, and I'm probably insane.


zakird:
Realistically it's going to be hard to review this given the length of this PR. I will read through it, but to verify this, please pull out a random subset of ten million certificates from Censys and make sure that they validate in the same manner as the old code, and in the cases that they do not, that you understand why they don't validate the same and that the new code has the correct value.

dadrian:
This is now ready for "final" review. A random sample of 100K certs had only 12 differences, each of which has been accounted for in https://docs.google.com/spreadsheets/d/1NzeYWx39wUuyjNHC5DaS9d1Idg4dbB5ryVi7VsjRnl8/edit?usp=sharing.

@zakird @cdzombak @umbernhard @ajholland @zzma

PR #86, opened 2017-08-20
Add GivenName and Surname subject fields

kumarde:
These are required to be checked by the BRs, so I'm adding them here.


kumarde:
@zakird I added them to the JSOn as well, which I neglected to do earlier.

PR #99, opened 2017-11-03
Add ParsedDomainName map to the Certificat struct to allow speeding up zlint

szank:
I have added a `ParsedDomainsMap map[string]ParsedDomainName` to the certificate struct. 
It will hold parsed DNS names (either CommonName or SAN DNS), and a parsing error, where the key is the domain name string. 
By parsing the domain names once and reusing in in multiple zlint linters the slowes linters could be sped up tremendously.

### DIFF
--- a/x509/x509.go
+++ b/x509/x509.go
@@ -42,6 +42,7 @@ type pkixPublicKey struct {
 	BitString asn1.BitString
 }
 
+// ParsedDomainName is a structure holding a parsed domain name (CommonName or DNS SAN) and a parsing error.
 type ParsedDomainName struct {
 	Domain     *publicsuffix.DomainName
 	ParseError error
--- a/x509/x509.go
+++ b/x509/x509.go
@@ -31,6 +31,8 @@ import (
 
 	"github.com/zmap/zcrypto/ct"
 	"github.com/zmap/zcrypto/x509/pkix"
+
+	"github.com/weppos/publicsuffix-go/publicsuffix"
 )
 
 // pkixPublicKey reflects a PKIX public key structure. See SubjectPublicKeyInfo
@@ -38,6 +40,11 @@ import (
 type pkixPublicKey struct {
 	Algo      pkix.AlgorithmIdentifier
 	BitString asn1.BitString
+}
+
+type ParsedDomainName struct {
+	Domain     *publicsuffix.DomainName
+	ParseError error
 }
 
 // ParsePKIXPublicKey parses a DER encoded public key. These values are
@@ -628,6 +635,11 @@ type Certificate struct {
 
 	// CT
 	SignedCertificateTimestampList []*ct.SignedCertificateTimestamp
+
+	// Used to speed up the zlint checks. Populated by zlint.
+	// Each map key is a potential domain name ( CommonName or DNS SAN)
+	// Each corresponding key is a structure holding a parsed domain name and a parsing error
+	ParsedDomainsMap map[string]ParsedDomainName
 }
 
 // SubjectAndKey represents a (subjecty, subject public key info) tuple.
@@ -1551,6 +1563,7 @@ func parseCertificate(in *certificate) (*Certificate, error) {
 		//	return out, UnhandledCriticalExtension{e.Id}
 		//}
 	}
+
 	return out, nil
 }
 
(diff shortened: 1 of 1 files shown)


zakird:
It might be best to have a method on certificate object that populates this map instead of relying on an external library to populate it.

szank:
Ah, it seems that I misunderstood your comment (https://github.com/zmap/zlint/issues/178#issuecomment-341682814). 
I assumed it should be ok for the zlint package to lazily populate the map when the linters are run. Will add a method to the Certificate struct then.

dadrian:
I'll give this a pass once that change has been made.

x509/x509.go
```diff
@@ -2129,6 +2143,40 @@ func (c *Certificate) CreateCRL(rand io.Reader, priv interface{}, revokedCerts [
...
+		if DNSName == "" {
+			continue
+		}
+		parsedDomain, parseError = publicsuffix.ParseFromListWithOptions(publicsuffix.DefaultList,
+			DNSName,
+			&publicsuffix.FindOptions{IgnorePrivate: true, DefaultRule: publicsuffix.DefaultRule})
+
+		c.ParsedDomainsMap[DNSName] = ParsedDomainName{
+			Domain:     parsedDomain,
+			ParseError: parseError,
```
dadrian: Can we unify the handling between DNSName and CommonName, similar to https://github.com/zmap/zlint/blob/2c8be8ad1a769ec3544da976aff3a0310f8b0628/util/fqdn.go#L81

szank:
Actually, it might be better, to make the `ParsedDomainsMap` private, and create a function
`func (c *Certificate) ParseDomainName() (*publicSuffix.DomainName, error)` 
that would lazily initialise the  `ParsedDomainsMap` and populate it with the parsed domain names. 

I can also squash the commits and create a new, clean PR if you guys prefer it.

zakird:
I agree that sounds a bit cleaner. Up to you on how you want to submit PR.
Doesn't matter much to us.

szank:
Hi, 
After some internal discussions, I have reworked the code. Now it should be obvious at the first glance that the parsed domains are computer from the DNS names and subject CommonName. 

I think this is a cleaner approach.

x509/x509.go
```diff
@@ -628,6 +636,11 @@ type Certificate struct {
 
 	// CT
 	SignedCertificateTimestampList []*ct.SignedCertificateTimestamp
+
+	// Used to speed up the zlint checks. Populated by the GetParsedDNSNames method.
+	parsedDNSNames []ParsedDomainName
+	// Used to speed up the zlint checks. Populated by the GetParsedCommonName method
+	parsedCommonName *ParsedDomainName
```
zakird: I'm a bit surprised that one of these is an array of objects and one is a pointer. Is there a reason not to just store both as objects on certificate?

szank: Both fields needs to be nillable. 
`parsedDNSNames` is a slice, because the DNS SANs are a slice. So storing parsed DNS SANs in a slice makes sense. 

For the CommonName, I need to discern between common name being an empty string and it not being parsed yet. This could be done by checking is parsedCommonName.ParsedDomain is nil or not, but in general using a pointer to the whole ParsedDomainName struct seems cleaner.

szank: Also, as the linters first try to parse the common name, and then proceed to SAN DNS names if parsing of the common name succeeds, it was a reasonable choice to split those two concerns into separate fields in the cert struct and separate function calls.

x509/x509.go
```diff
@@ -40,6 +41,13 @@ type pkixPublicKey struct {
 	BitString asn1.BitString
 }
 
+// ParsedDomainName is a structure holding a parsed domain name (CommonName or DNS SAN) and a parsing error.
+type ParsedDomainName struct {
+	DomainString string
+	ParsedDomain *publicsuffix.DomainName
```
zakird: Does this also need to be a pointer?

szank: Not necessarily. The  `publicsuffix.DomainFromListWithOptions` returns a pointer, so I kept is as a pointer. I could run some benchmarks to see if it makes sense to dereference it here if you wish.

zakird:
Sounds good. I've merged.

PR #100, opened 2017-11-15
TLSCertsOnly Option

jcrowgey:
This PR is concomitant with my PR to zgrab (which implements the option).

This option allows clients to abort the connection after receiving the server_certificates message.  This allows offline certificate parsing and validation (which is expensive), freeing up resources and allowing the scanning host to move faster.

### DIFF
--- a/tls/common.go
+++ b/tls/common.go
@@ -496,6 +496,10 @@ type Config struct {
 	// material from the returned config will be used for session tickets.
 	GetConfigForClient func(*ClientHelloInfo) (*Config, error)
 
+	// TLSCertsOnly is used to cause a client to close the TLS connection
+	// as soon as the server's certificates have been received
+	TLSCertsOnly bool
+
 	// mutex protects sessionTicketKeys and originalConfig.
 	mutex sync.RWMutex
 	// sessionTicketKeys contains zero or more ticket keys. If the length
--- a/tls/handshake_client.go
+++ b/tls/handshake_client.go
@@ -515,9 +515,10 @@ func (c *Conn) clientHandshake() error {
 		if err := hs.doFullHandshake(); err != nil {
 			return err
 		}
-
-		return nil
-
+        if c.config.TLSCertsOnly {
+            // All done
+		    return nil
+        }
 		if err := hs.establishKeys(); err != nil {
 			return err
 		}
@@ -612,7 +613,6 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 			c.verifiedChains, validation, err = certs[0].ValidateWithStupidDetail(opts)
 			c.handshakeLog.ServerCertificates.addParsed(certs, validation)
 
-
 			// If actually verifying and invalid, reject
 			if !c.config.InsecureSkipVerify {
 				if err != nil {
@@ -670,7 +670,6 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		if err != nil {
 			return err
 		}
-
 	}
 
 	// If we don't support the cipher, quit before we need to read the hs.suite
--- a/tls/handshake_client.go
+++ b/tls/handshake_client.go
@@ -515,6 +515,9 @@ func (c *Conn) clientHandshake() error {
 		if err := hs.doFullHandshake(); err != nil {
 			return err
 		}
+
+		return nil
+
 		if err := hs.establishKeys(); err != nil {
 			return err
 		}
@@ -583,6 +586,11 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 
 		c.handshakeLog.ServerCertificates = certMsg.MakeLog()
 
+		if c.config.TLSCertsOnly {
+		  // short circuit!
+  		  return nil
+		}
+
 		if !invalidCert {
 			opts := x509.VerifyOptions{
 				Roots:         c.config.RootCAs,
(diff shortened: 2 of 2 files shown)


tls/handshake_client.go
```diff
@@ -515,6 +515,10 @@ func (c *Conn) clientHandshake() error {
 		if err := hs.doFullHandshake(); err != nil {
 			return err
 		}
+        if c.config.TLSCertsOnly {
+            // All done
+		    return nil
```
timothycrosley: This indentation looks a little off, otherwise everything LGTM

tls/common.go
```diff
@@ -496,6 +496,10 @@ type Config struct {
 	// material from the returned config will be used for session tickets.
 	GetConfigForClient func(*ClientHelloInfo) (*Config, error)
 
+	// TLSCertsOnly is used to cause a client to close the TLS connection
+	// as soon as the server's certificates have been received
+	TLSCertsOnly bool
```
dadrian: Given that this is already in the `Config` struct in the `tls` package, I think we can just call this `CertsOnly`.

jcrowgey:
Hi as well.  

I've updated the PR in accordance with your requested change in https://github.com/zmap/zcrypto/pull/100/commits/9b39b1586248e55f610c5c1fc6ac508d8a318698.  I also updated the related PR for zgrab https://github.com/zmap/zgrab/pull/297

Cheers!

jcrowgey:
It's been a while, is it okay to bump this PR discussion?  If there's other things I ought to do to have this merged, I'd love to hear more.  Cheers.

tls/common.go
jcrowgey: renamed in https://github.com/zmap/zcrypto/pull/100/commits/9b39b1586248e55f610c5c1fc6ac508d8a318698

tls/handshake_client.go
```diff
@@ -515,6 +515,10 @@ func (c *Conn) clientHandshake() error {
 		if err := hs.doFullHandshake(); err != nil {
 			return err
 		}
+		if c.config.CertsOnly {
+			// All done
+			return nil
```
dadrian: This should return a package-level error that clients can compare against, similar to `io.EOF`. Something like `ErrCertsOnly`, rather than `nil`.

dadrian:
Sorry for making you wait so long for feedback :(

tls/handshake_client.go
jcrowgey: @dadrian that makes sense to me.  Thanks for the feedback, I'll see if I can generate a new commit with that change.

jcrowgey: Just to make sure I'm on the same page, I think the idea is that the short circuit in line 592 should have something like

```
return errors.New("tls: handshake abandoned per TLSCertsOnly option: " + certsOnlyErr.Error())
```

And then the lines in 518 -- 520 would just be removed (because 515,516 already passes along errors from hs.doFullHandshake().

Is that what you're thinking?

jcrowgey:
@dadrian I created ErrCertsOnly in common.go, however, I noticed some other error types being created locally to where they were returned.  Let me know if I've done this in a way which is acceptable.

Secondly, I tried to add a test to handshake_client_test.go for the certs only option (planning on just testing that doFullHandshake returns the proper error type.  However, I had some difficulty because it seemed that at some point handshake_server started trying to utilize resume the connection, and ended up failing with error EOF.  I'd like to add a proper test, so if you have any advice on this, let me know.

Finally, I considered having the short circuit call conn.sendAlert(alertCloseNotify).  The idea here is that if we send a close_notify message, servers shouldn't record errors about abortive connections.  If you have any thoughts about whether that's the right approach, and if so, which part of the codebase should listen for ErrCertsOnly and send close_notify, please share!

Cheers.

tls/handshake_client.go
```diff
@@ -513,6 +513,9 @@ func (c *Conn) clientHandshake() error {
 		}
 	} else {
 		if err := hs.doFullHandshake(); err != nil {
+            if err == ErrCertsOnly {
+                c.sendAlert(alertCloseNotify)
+            }
```
dadrian: Indentation if off here.

jcrowgey: Sorry about that.  My typical tabstop and shiftwidth are interacting with this project in a kooky way where things look good to me in vi but are actually messed up.  I believe this is fixed in https://github.com/zmap/zcrypto/pull/100/commits/45bd5336e1b95a543b30e7c71200eacf30e8d0b4

jcrowgey:
Great, I'm looking forward to getting this merged :)

I've also updated my PR for zgrab which makes this tls option avaiable (https://github.com/zmap/zgrab/pull/297) to listen for `ErrCertsOnly` when scanning with `--tls-certs-only`.

My plan is to play around with finishing up the handshake but skipping parsing and validation of the certificates.  Some initial testing I did suggests that parsing and validation is the real bottleneck, and skipping the rest of the handshake triggers messages in server logs (depending on the server software and the log level) which may be problematic.

dadrian:
Thanks for putting up with how slow we were on this!

jcrowgey:
@dadrian no problem at all.  Thanks very much to you guys for the open-source code.  I've been using these features that I'm offering in these PRs, I'm happy to give back to the project if at all possible.

PR #103, opened 2018-01-08
Finish handshake skip validation

jcrowgey:
This PR is a follow up on PR #100, which introduced the ability to early-abort the handshake.  One drawback of the early abort is that it can create noise in the logs of servers with verbose logging.

The option introduced in this PR "SkipValidation", allows a user to skip all validation and parsing of the certificate chain (for speed, the use case is for post-processing offline), while still finishing the handshake cleanly.

My early testing suggests this is about 2x slower than --certs-only, but still about 4--5x faster than the default behavior.

### DIFF
--- a/tls/common.go
+++ b/tls/common.go
@@ -496,6 +496,10 @@ type Config struct {
 	// material from the returned config will be used for session tickets.
 	GetConfigForClient func(*ClientHelloInfo) (*Config, error)
 
+	// TLSCertsOnly is used to cause a client to close the TLS connection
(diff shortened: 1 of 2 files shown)


dadrian:
@jcrowgey Can you merge master back into this, now that #100 landed?

jcrowgey:
This PR has a bug.  When the certificate is invalid, zgrap crashes.  I will close this PR and generate a new one when I've got the details straightened out.

PR #105, opened 2018-01-17
tls: add tls.Config option to disable buffered handshakes

justinbastress:
Adds a `DontBufferHandshakes` tls.Config option. Setting this to `true` reverts to the prior behavior, where each handshake packet is sent in its own `conn.Write()`.

## How to Test

The included unit tests do work; you can also use Wireshark etc to confirm that more packets get sent in the handshake when `DontBufferHandshakes = true`.

The only place I have seen different behavior is with MSSQL, but getting to that point is non-trivial.

## Notes & Caveats

I didn't see any other examples of a default-true config option; the double negative `DontBufferHandshakes = false` isn't ideal, but it at least seemed clearer than e.g. `SendUnbufferedHandshakes = false`.

Also ensures that the buffers get flushed on a panic.

## Issue Tracking

Pre-requisite for https://github.com/zmap/zgrab2/issues/18

### DIFF
--- a/tls/common.go
+++ b/tls/common.go
@@ -500,6 +500,9 @@ type Config struct {
 	// as soon as the server's certificates have been received
 	CertsOnly bool
 
+	// DontBufferHandshakes causes Handshake() to act like older versions of the go crypto library, where each TLS packet is sent in a separate Write.
+	DontBufferHandshakes bool
+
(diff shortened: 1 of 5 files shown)


zakird:
@justinbastress this looks to fine to merge once conflicts are fixed up.

Issue #106, opened 2018-01-26
Zcrypto refuses to comile with go 1.10rc1 on macos


szank:
Hi, I have been testing the go@1.10rc and discovered that this package doesn't compile: 
I am getting:
`vendor/github.com/zmap/zcrypto/ct/x509/root_darwin.go:73: cannot use nil as type _Ctype_CFDataRef in assignment`

After a quick glance I have discovered that zcrypto/ct/x509 is quite outdated compared to the https://github.com/google/certificate-transparency-go and the baseline x509 package. The same thing for zcrypto/ct/asn1 and possibly zcrypto/x509. 

Updating the offending functions should be easy enough,  and I could do that, but I tried to merge github.com/google/certificate-transparency-go into zcrypto and it turned out bit tricky. 
There are small changes sprinkled here and there, and without knowing what and where have been modified it is very easy to miss something. And not all is wrapped in `// Begin CT changes`, `// End CT changes`. 

I can open a PR to just update the offending parts, so there is no nasty surprise when other users of this repo jump to go 1.10, but I'd suggest an general update to the x509/asn1 packages.

adamdecaf:
I was able to get around with this smaller fix. PR'ing as I wasn't able to find your change specific to those couple of errors. (Rather than updating the x509 forks) 

https://github.com/zmap/zcrypto/pull/109

szank:
This is solved no. Zcrypto works on go1.10 and previous versions.

Issue #9, opened 2017-03-10
CT package should use zcrypto/x509


dadrian:
The CT package currently contains it's own fork of `crypto/x509` and `crypto/asn1`. It's not entirely clear why these are needed. At the very least, we should make sure whatever changes in the `ct/x509` module are in `zcrypto/x509`, and try to remove that dependency. However, this will create a circular dependency between `zcrypto/x509` and `zcrypto/ct`. So we'll then have to figure how what the `ct` module needs any form of `x509` anyway, and possible extract that to a separate package.

Another possible solution might be to make it `ct` subpackage of `zcrypto/x509`.

dadrian:
I took a crack at this the other day. This is non-trivial. The `zcrypto/ct` code from Google is poorly organized, and tightly coupled to it's use of `zcrypto/ct/x509`, and there's not a clear division between what goes in a CT log, and what CT info goes in a certificate. I started splitting it up by moving all SCT-related code into our x509 library, so that we could get rid of the CT x509 library. This is also annoying because of the massive `serialization.go`, which defines a bunch of interrelated serialization functions that make it difficult to cut the package down the middle.

So in summary:
- It's entirely possible to remove `zcrypto/ct/x509`
- There will have to be some split between `zcrypto/ct` and `zcrypto/x509/ct` in order to avoid circular dependencies.
- Splitting `zcrypto/ct` in two is non-trivial.
- Whoever works on this will have to spend a decent amount of time getting familiar with the implementation of `zcrypto/ct` in order to do this cleanly. My first attempt to do this was largely copy-paste based, and got complicated fast. It needs a bit more methodology to do cleanly than copy paste, change names, fix compile errors. Some functions will need to be split, changed, defined on different objects, etc.

zakird:
I'm not sure this is worth it. We need to just scrap our entire CT client implementation and write a fresh one. As part of this, we should probably just write a clean client. We need to introduce new, fundamental pieces of functionality (e.g., verifying log entries), and so I'm a bit doubtful we'd want to use the existing Google one anyway.

PR #108, opened 2018-02-06
Update x509 to latest 1.9, keeping modifications and simplify ct

tadukurow:
This is a large PR that aims to separate some of the stdlib imported code from the modifications made by zcrypto. Introducing this separation should make it easier to track upstream and in many cases simplify some code, as the stdlib asn1 and x509 libraries have evolved quite a bit. 
It solves or facilitates https://github.com/zmap/zcrypto/issues/1, https://github.com/zmap/zcrypto/issues/9, https://github.com/zmap/zcrypto/issues/106

It also breaks the dependency of the x509 and tls package on the ct package. The reasoning is that the ct package contains it's own x509 and is difficult to extract and upgrade. So the code necessary to define and Deserialize SCTs has been moved to x509/ct.
This does mean upstream packages would need to update their import paths. 
Personally I think the ct package should not be part of the zcrypto repo as it introduces a lot of problems when trying to track upstream for asn1 and x509. if ct was removed as hinted in https://github.com/zmap/zcrypto/issues/9, the ct deserialization code used in tls and x509 code go back to the normal import path and updates would be needed. But I don't know enough about the usage patterns of CT within zmap.

First I determined that the original x509 port seems to be from around 2015, not sure what incremental updates were done. Unfortunately no changelog with revisions was documented. 
Then I updated it to latest 1.9 release (documented in README.md), preserving all modifications in files called $name_modified.go.
Test files weren't updated or modified and are passing. 
I left some notes where it seems zcrypto could align with the stdlib with some updates. 

This work should also allow for an easy upgrade path to 1.10 and newer versions.


justinbastress:
@tadukurow we are still looking into the best way to move forward here with regard to upstream golang crypto changes.

tadukurow:
@justinbastress thanks for the update. 

I'm happy to change anything on the PR if necessary. 

Also would you be able to share anything on the direction you're debating/planning to take?

tadukurow:
Hi is there any update on this? 

We're looking to move to 1.10 as it has some features we need. 
Would a PR that simply makes the build pass on 1.10 and 1.9 as suggested on #109, be considered at this point? 
https://github.com/google/certificate-transparency-go/commit/391726f8973ddb3521ab1a462bd55a35c6f67f0b

justinbastress:
@tadukurow this should go in as soon as I find the right way to pin other apps (e.g. zgrab) to the existing version.

In the meantime -- would you be able to share the context in which you are using zcrypto? e.g. just as a dependency for zlint, or something more involved?

tadukurow:
@justinbastress that's good news. 

we use zcrypto strictly as a dependency for zlint, but would like to keep contributing to zlint like the previous work by @szank.

PR #109, opened 2018-02-18
don't set C.CFDataRef explicitly to nil for darwin

adamdecaf:
A much smaller fix to https://github.com/zmap/zcrypto/issues/106. I was trying out `zcertificate` and ran into the same problem.


szank:
Please have a look at google CT repo: 
https://github.com/google/certificate-transparency-go/commit/391726f8973ddb3521ab1a462bd55a35c6f67f0b

The code in question was copied from there, and they have a fix for this issue that works both on go1.10 and go1.9.

adamdecaf:
Thanks. Are y'all planning to update from that repo again?

PR #110, opened 2018-02-20
Add InsecureSkipValidation option

justinbastress:
Follows https://github.com/zmap/zcrypto/pull/103, except moves the check for `if invalidCert` out of the `if !c.InsecureSkipValidation` block, so that the handshake is not attempted with a certificate that fails to parse (which is `invalid` in the sense of not being valid X.509, not the sense of not successfully path-validated).

**Note**: The original comments on `InsecureSkipValidation` indicated that the certificate was not even parsed -- this is not accurate, since the certificate must be parsed in order to complete the handshake.  This has been updated to reflect that here no validation is attempted.

### DIFF
--- a/tls/common.go
+++ b/tls/common.go
@@ -389,6 +389,11 @@ type Config struct {
 	// This should be used only for testing.
 	InsecureSkipVerify bool
 
+	// InsecureSkipValidation, is like InsecureSkipVerify is used to skip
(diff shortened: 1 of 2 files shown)


justinbastress:
Closing per comment at https://github.com/zmap/zgrab/pull/303#issuecomment-367424722.

PR #122, opened 2018-04-16
Add zcrypto schema

justinbastress:
**Depends on https://github.com/zmap/zschema/compare/master...subrecord-types.**

For an example of what this looks like in use, see https://github.com/zmap/zgrab2/compare/master...feature/schemaUpdate.

Next step would be merging this into ztag (having it use the `SubRecordType`s like `ParsedCertificate` from zcrypto/schema.py instead of the `SubRecord`s like `zgrab_parsed_certificate`).


justinbastress:
@cdzombak agreed; the only place I foresee overlap is in the TLS handshake, since the ztag version is significantly more stripped-down.

PR #123, opened 2018-04-19
tls: fix errors found by vet

functionary:
This pull request was made with an automated tool.

The suggested change fixes one or more problems discovered by "go vet".

See https://github.com/functionary/functionary for more details on
the @functionary GitHub user.

### DIFF
--- a/tls/handshake_messages.go
+++ b/tls/handshake_messages.go
@@ -556,7 +556,7 @@ func (m *clientHelloMsg) unmarshal(data []byte) bool {
 			if length < 3 {
 				return false
 			}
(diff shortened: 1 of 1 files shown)


adamdecaf:
@justinbastress Do you think https://github.com/golang/go/issues/25087 would apply here? That might be a useful test case for the compiler.

rasky:
No it wouldn’t. That issue is about non-constant shifts where the compiler has to go through some hurdles because shifting by an amount larger than word size produces zero in Go but not on x86 processors.

Shifting a byte by 8 (or 100) is correct Go code (yielding 0) and the compiler is expected to accept it. This is why there’s vet after all.

PR #124, opened 2018-04-25
schemas.zcrypto egg

justinbastress:
Add setup.py to allow `schemas.zcrypto` to be imported as an egg (`git+https://github.com/zmap/zcrypto#egg=schemas.zcrypto`)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,25 @@
+import os
+from setuptools import setup, find_packages
+
+here = os.path.abspath(os.path.dirname(__file__))
+
+setup(
+  name="schemas.zcrypto",
+  version="0.0.1",
+  description="ZSchema definitions for zcrypto's JSON output.",
+  classifiers=[
+    "Programming Language :: Python",
+    "Natural Language :: English"
+  ],
+  author="ZMap Team",
+  author_email="zmap-team@umich.edu",
+  url="https://github.com/zmap/zcrypto",
+  keywords="zmap censys ztag internet-wide scanning",
+  packages=["schemas"],
+  include_package_data=True,
+  zip_safe=False,
+  install_requires = [
+    "zschema",
+  ],
+  # package_data={"ztag":["devices/*",]},
+)
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import setup, find_packages
 here = os.path.abspath(os.path.dirname(__file__))
 
 setup(
-  name="schemas.zcrypto",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -0,0 +1,25 @@
...
+setup(
+  name="schemas.zcrypto",
+  version="0.0.1",
+  description="ZSchema definitions for zcrypto's JSON output.",
+  classifiers=[
+    "Programming Language :: Python",
+    "Natural Language :: English"
+  ],
+  author="ZMap Team",
+  author_email="zmap-team@umich.edu",
```
zakird: this should move to team@zmap.io

setup.py
```diff
@@ -0,0 +1,25 @@
+import os
+from setuptools import setup, find_packages
+
+here = os.path.abspath(os.path.dirname(__file__))
+
+setup(
+  name="schemas.zcrypto",
```
zakird: It's probably good not to have a period in a package name given that's how you go from a package to a module.

setup.py
```diff
@@ -0,0 +1,25 @@
...
+  description="ZSchema definitions for zcrypto's JSON output.",
+  classifiers=[
+    "Programming Language :: Python",
+    "Natural Language :: English"
+  ],
+  author="ZMap Team",
+  author_email="zmap-team@umich.edu",
+  url="https://github.com/zmap/zcrypto",
+  keywords="zmap censys ztag internet-wide scanning",
+  packages=["schemas"],
```
zakird: It might be more natural to have a package name that includes "zcrypto" in it since I could see having multiple ones of these to import and this will be the package that's imported.

setup.py
justinbastress: Will do. I copied this from ztag, so presumably that needs to be updated too?

setup.py
justinbastress: I'll see if I can find an example of this (e.g. multiple different components dropping modules under `http.*`); right now, both `zgrab2` and `zcrypto` exist under `schemas` and it seems to "work", but that was just an empirical finding.

setup.py
justinbastress: gotcha. I had it set up to match the package name, but that doesn't seem to be necessary. I'll change this to `zcypto-schema` (and `zgrab2-schema`).

setup.py
zakird: Yeah, we should update it there too.

justinbastress:
@zakird I think that addresses all of your comments.

PR #132, opened 2018-06-01
re-add exclude=bigquery to EKU and CP extensions

justinbastress:


justinbastress:
actually -- now that the bogus records are being pruned, do we still want to exclude these?

PR #139, opened 2018-07-10
Fix compilation on OS X on latest Go

### LINKED ISSUE #121: ct/x509: Package fails to build on Darwin with Go 1.10.1
`zgrab` fails to build on OS X running Go 1.10.1:

```sh
$ go get -u github.com/zmap/zgrab
# github.com/zmap/zgrab/vendor/github.com/zmap/zcrypto/ct/x509
vendor/github.com/zmap/zcrypto/ct/x509/root_darwin.go:73: cannot use nil as type _Ctype_CFDataRef in assignment
```

Related: https://github.com/google/certificate-transparency-go/issues/131
Fix seems to be in https://github.com/google/certificate-transparency-go/pull/155

dadrian:
This copy/pastes all the root*.go files from x509/ to ct/x509/, in order
to fix the cgo errors being exposed in the latest version of Go.

This is another reason that it'd be great if the CT library didn't have
to embed it's own x509 library.

Fixes #121


zakird:
ugh, we should get rid of this code. :|

PR #142, opened 2018-07-19
New Revocation Checking Functionality

casebenton:
Added two new packages for checking revocation status of certificates through CRL and OCSP. Packages include functionality for constructing and parsing messages over these two protocols.

### DIFF
--- a/x509/revocation/crl/crl.go
+++ b/x509/revocation/crl/crl.go
@@ -1,0 +1,143 @@
+package crl
+
+import (
+	"encoding/asn1"
+	"strconv"
+	"time"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zcrypto/x509/pkix"
+)
+
+// RevocationReasonCode - status codes that explain revocation reason see RFC 5620
+type RevocationReasonCode int
+
+const (
+	Unspecified          RevocationReasonCode = 0
+	KeyCompromise        RevocationReasonCode = 1
+	CACompromise         RevocationReasonCode = 2
+	AffiliationChanged   RevocationReasonCode = 3
+	Superseded           RevocationReasonCode = 4
+	CessationOfOperation RevocationReasonCode = 5
+	CertificateHold      RevocationReasonCode = 6
+	// STATUS CODE 7 IS NOT USED
+	RemoveFromCRL      RevocationReasonCode = 8
+	PrivilegeWithdrawn RevocationReasonCode = 9
+	AACompromise       RevocationReasonCode = 10
+)
+
+func (r RevocationReasonCode) String() string {
+	switch r {
+	case Unspecified:
+		return "Unspecified"
+	case KeyCompromise:
+		return "Key Compromise"
+	case CACompromise:
+		return "CA Compromise"
+	case AffiliationChanged:
+		return "Affiliation Changed"
+	case Superseded:
+		return "Superseded"
+	case CessationOfOperation:
+		return "Cessation Of Operation"
+	case CertificateHold:
+		return "Certificate Hold"
+		// STATUS CODE 7 IS NOT USED
+	case RemoveFromCRL:
+		return "Remove From CRL"
+	case PrivilegeWithdrawn:
+		return "PrivilegeWithdrawn"
+	case AACompromise:
+		return "AA Compromise"
+	default:
+		return "Unknown revocation reason code: " + strconv.Itoa(int(r))
+	}
+}
+
+var (
+	crlNumberExtensionOID        = asn1.ObjectIdentifier{2, 5, 29, 20}
+	revocationReasonExtensionOID = asn1.ObjectIdentifier{2, 5, 29, 21}
+	invalidityDateExtensionOID   = asn1.ObjectIdentifier{2, 5, 29, 24}
+)
+
+type crlNumberExtension struct {
+	ID        asn1.ObjectIdentifier
+	Critical  bool `asn1:"optional"`
+	CRLNumber int
+}
+
+// TODO: handle additional CRL Extensions
+// 2.5.29.21,cRLReason
+// id-ce-cRLReasons OBJECT IDENTIFIER ::= { id-ce 21 }
+// -- reasonCode ::= { CRLReasonCode }
+
+//2.5.29.24,invalidityDate
+// id-ce-invalidityDate OBJECT IDENTIFIER ::= { id-ce 24 }
+// -- InvalidityDate ::=  GeneralizedTime
+
+// RevocationData - information on whether a certificate has been
+// revoked by a specified CRL, and information on the CRL
+type RevocationData struct {
+	CRLSignatureAlgorithm         x509.SignatureAlgorithm
+	CRLSignatureValue             []byte
+	Version                       int `asn1:"optional,default:0"`
+	Issuer                        pkix.Name
+	ThisUpdate                    time.Time
+	NextUpdate                    time.Time `asn1:"optional"`
+	CRLExtensions                 ListExtensionData
+	UnknownCRLExtensions          []pkix.Extension `asn1:"tag:0,optional,explicit"`
+	IsRevoked                     bool
+	RevocationTime                time.Time
+	CertificateEntryExtensions    RevokedCertExtensionData
+	RawCertificateEntryExtensions []pkix.Extension `asn1:"optional"`
+}
+
+// ListExtensionData - Data from optional, non-critical pkix.CertificateList extensions
+type ListExtensionData struct {
+	CRLNumber int
+	AuthKeyID x509.SubjAuthKeyId `json:"authority_key_id,omitempty"`
+}
+
+// RevokedCertExtensionData - Data from optional, non-critical pkix.RevokedCertificate extensions
+type RevokedCertExtensionData struct {
+	Reason         string `json:"revocation_reason,omitempty"`
+	invalidityDate time.Time
+}
+
+func gatherListExtensionInfo(certList *pkix.CertificateList, ret *RevocationData) {
+	for _, extension := range certList.TBSCertList.Extensions {
+		if extension.Id.Equal(crlNumberExtensionOID) {
+			var ext crlNumberExtension
+			asn1.Unmarshal(extension.Value, &ext.CRLNumber)
+			ret.CRLExtensions.CRLNumber = ext.CRLNumber
+		} else {
+			ret.UnknownCRLExtensions = append(ret.UnknownCRLExtensions, extension)
+		}
+	}
+}
+
+// CheckCRLForCert - parses through a given CRL and to see if a given certificate
+// is present, and returns data on the revocation and CRL in general
+func CheckCRLForCert(certList *pkix.CertificateList, cert *x509.Certificate) (*RevocationData, error) {
+	ret := &RevocationData{
+		CRLSignatureAlgorithm: x509.GetSignatureAlgorithmFromAI(certList.SignatureAlgorithm),
+		CRLSignatureValue:     certList.SignatureValue.Bytes,
+		Version:               certList.TBSCertList.Version,
+		ThisUpdate:            certList.TBSCertList.ThisUpdate,
+		NextUpdate:            certList.TBSCertList.NextUpdate,
+		IsRevoked:             false,
+	}
+	ret.Issuer.FillFromRDNSequence(&certList.TBSCertList.Issuer)
+
+	gatherListExtensionInfo(certList, ret)
+
+	revokedCerts := certList.TBSCertList.RevokedCertificates
+	for i := range revokedCerts {
+		if revokedCerts[i].SerialNumber == cert.SerialNumber {
+			ret.IsRevoked = true
+			ret.RevocationTime = revokedCerts[i].RevocationTime
+			break
+		}
+	}
+	return ret, nil
+}
--- a/x509/revocation/ocsp/ocsp.go
+++ b/x509/revocation/ocsp/ocsp.go
@@ -1,0 +1,422 @@
+package ocsp
+
+import (
+	"crypto"
+	"encoding/asn1"
+	"errors"
+	"math/big"
+	"strconv"
+	"time"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zcrypto/x509/pkix"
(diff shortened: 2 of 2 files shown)


x509/revocation/crl/crl.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func gatherListExtensionInfo(certList *pkix.CertificateList, ret *RevocationData) {
+	for _, extension := range certList.TBSCertList.Extensions {
+		if extension.Id.Equal(crlNumberExtensionOID) {
+			var ext crlNumberExtension
+			asn1.Unmarshal(extension.Value, &ext.CRLNumber)
+			ret.CRLExtensions.CRLNumber = ext.CRLNumber
+		} else {
+			ret.UnknownCRLExtensions = append(ret.UnknownCRLExtensions, extension)
```
justinbastress: Would it make sense to separately track unknown *critical* extensions?

x509/revocation/crl/crl.go
```diff
@@ -0,0 +1,143 @@
...
+		ThisUpdate:            certList.TBSCertList.ThisUpdate,
+		NextUpdate:            certList.TBSCertList.NextUpdate,
+		IsRevoked:             false,
+	}
+	ret.Issuer.FillFromRDNSequence(&certList.TBSCertList.Issuer)
+
+	gatherListExtensionInfo(certList, ret)
+
+	revokedCerts := certList.TBSCertList.RevokedCertificates
+	for i := range revokedCerts {
```
justinbastress: Would it make sense to store this internally as a map for quick lookup?

x509/revocation/ocsp/ocsp.go
```diff
@@ -0,0 +1,422 @@
...
+// Request represents an OCSP request. See RFC 6960.
+type Request struct {
+	HashAlgorithm  crypto.Hash
+	IssuerNameHash []byte
+	IssuerKeyHash  []byte
+	SerialNumber   *big.Int
+}
+
+// https://tools.ietf.org/html/rfc2560#section-4.1.1
+type ocspRequest struct {
```
justinbastress: It seems like these would be helpful to have exported -- was there a reason to make them private?

x509/revocation/ocsp/ocsp.go
```diff
@@ -0,0 +1,422 @@
...
+		Algorithm pkix.AlgorithmIdentifier
+		PublicKey asn1.BitString
+	}
+
+	if _, err := asn1.Unmarshal(issuer.RawSubjectPublicKeyInfo, &publicKeyInfo); err != nil {
+		return nil, err
+	}
+
+	h.Write(publicKeyInfo.PublicKey.RightAlign())
+	issuerKeyHash := h.Sum(nil)
```
justinbastress: It could be handy to have this in its own API, e.g. `ocsp.GetKeyHash(cert *x509.Certificate)`

x509/revocation/ocsp/ocsp.go
```diff
@@ -0,0 +1,422 @@
...
+
+	h.Reset()
+	h.Write(issuer.RawSubject)
+	issuerNameHash := h.Sum(nil)
```
justinbastress: Same as above -- `ocsp.GetNameHash(cert *x509.Certificate)`? I'm not sure what the best way to input the digest would be though.

x509/revocation/ocsp/ocsp.go
```diff
@@ -0,0 +1,422 @@
...
+	crypto.SHA384: asn1.ObjectIdentifier([]int{2, 16, 840, 1, 101, 3, 4, 2, 2}),
+	crypto.SHA512: asn1.ObjectIdentifier([]int{2, 16, 840, 1, 101, 3, 4, 2, 3}),
+}
+
+// ParseResponseForCert parses an OCSP response in DER form and searches for a
+// Response relating to cert. If such a Response is found and the OCSP response
+// contains a certificate then the signature over the response is checked. If
+// issuer is not nil then it will be used to validate the signature or embedded
+// certificate.
+//
```
justinbastress: Would it make sense to split this `ParseResponse`, `response.GetResponseForCert`, and `response.Validate(signer *x509.Certificate)`? A combined version (say, `ocsp.CheckOCSPResponse(derResponse []byte, cert *x509.Certificate, signer *x509.Certificate)` would be good too, but having the primitives seems important too.

x509/revocation/ocsp/ocsp.go
```diff
@@ -0,0 +1,422 @@
...
+		if rest, err = asn1.Unmarshal(rawResponderID.Bytes, &ret.ResponderKeyHash); err != nil || len(rest) != 0 {
+			err = errors.New("invalid responder key hash")
+			return nil, err
+		}
+	default:
+		err = errors.New("invalid responder id tag")
+		return nil, err
+	}
+
+	if len(basicResp.Certificates) > 0 {
```
justinbastress: IIRC the CA and the ocsp signer either need to be the same certificate, or siblings...am I reading this wrong, or does this miss that check?

x509/revocation/crl/crl.go
```diff
@@ -0,0 +1,143 @@
...
+// -- reasonCode ::= { CRLReasonCode }
+
+//2.5.29.24,invalidityDate
+// id-ce-invalidityDate OBJECT IDENTIFIER ::= { id-ce 24 }
+// -- InvalidityDate ::=  GeneralizedTime
+
+// RevocationData - information on whether a certificate has been
+// revoked by a specified CRL, and information on the CRL
+type RevocationData struct {
+	CRLSignatureAlgorithm         x509.SignatureAlgorithm
```
justinbastress: This goes for all types -- it would be nice if these all had standard `json` fields, and if necessary, custom `MarshalJSON` / 'UnmarshalJSON` methods (say, for `RevocationReasonCode`, it could encode to the string value, but take either the int or string value when decoding)

x509/revocation/crl/crl.go
```diff
@@ -0,0 +1,143 @@
...
+	RemoveFromCRL      RevocationReasonCode = 8
+	PrivilegeWithdrawn RevocationReasonCode = 9
+	AACompromise       RevocationReasonCode = 10
+)
+
+func (r RevocationReasonCode) String() string {
+	switch r {
+	case Unspecified:
+		return "Unspecified"
+	case KeyCompromise:
```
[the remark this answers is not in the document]
casebenton: Good call, I'll switch it to camelCase as specified in the RFC.

x509/revocation/ocsp/ocsp.go
casebenton: No reason in particular to make them private - I just figured it would be a good approach to minimize the number of exported structs. I'll switch them over to public structs though if you think that is more appropriate in this situation.

x509/revocation/ocsp/ocsp.go
```diff
@@ -0,0 +1,422 @@
...
+	Success       ResponseStatus = 0
+	Malformed     ResponseStatus = 1
+	InternalError ResponseStatus = 2
+	TryLater      ResponseStatus = 3
+	// STATUS CODE 4 IS UNUSED IN OCSP
+	SignatureRequired ResponseStatus = 5
+	Unauthorized      ResponseStatus = 6
+)
+
+func (r ResponseStatus) String() string {
```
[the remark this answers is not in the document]
casebenton: Done, see https://github.com/zmap/zcrypto/pull/142/commits/8a2b23003d430dbad260c9de57ba9b3467c2632e

x509/revocation/ocsp/ocsp.go
```diff
@@ -0,0 +1,422 @@
...
+// SignatureRequired - Must Sign the Request
+// Unauthorized - Request Unauthorized
+// See https://tools.ietf.org/html/rfc6960#section-4.2.1
+const (
+	Success       ResponseStatus = 0
+	Malformed     ResponseStatus = 1
+	InternalError ResponseStatus = 2
+	TryLater      ResponseStatus = 3
+	// STATUS CODE 4 IS UNUSED IN OCSP
+	SignatureRequired ResponseStatus = 5
```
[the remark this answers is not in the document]
casebenton: Done, see https://github.com/zmap/zcrypto/pull/142/commits/8a2b23003d430dbad260c9de57ba9b3467c2632e

x509/revocation/ocsp/ocsp.go
casebenton: I believe that the signer must either be the exact key that signed the cert, or a key that was explicitly delegated to sign the response (and is provided with the response) https://tools.ietf.org/html/rfc6960#section-4.2.2.2. 

I moved this signature checking code into its own function and added commenting to make it more clear https://github.com/zmap/zcrypto/pull/142/commits/e02c563ed732d505ac262c522861302ae333a5be. I don't think that it misses any checks, how does it look now?

x509/revocation/ocsp/ocsp.go
casebenton: how does it look now with https://github.com/zmap/zcrypto/pull/142/commits/6f47c567b32330cfa177e31145a31ae0a1354e0a and https://github.com/zmap/zcrypto/pull/142/commits/e02c563ed732d505ac262c522861302ae333a5be? Do you think it should be broken up further?

x509/revocation/crl/crl.go
casebenton: That's a good idea. Added this with https://github.com/zmap/zcrypto/pull/142/commits/d367ce200be69c6c8990436ba81fdeb7237344d4

x509/revocation/crl/crl.go
casebenton: Good call, I added an optional argument to this function so that clients can pass in a map for quick lookups https://github.com/zmap/zcrypto/pull/142/commits/914bf40eb7e5a3bcd34362f6d7e65f776a1906b8

x509/revocation/ocsp/ocsp.go
justinbastress: This looks like the right balance between flexibility and being fiddly -- `ParseResponse` gives exactly the same information that you could get from parsing the ASN.1 yourself, in one line instead of 15.

x509/revocation/ocsp/ocsp.go
justinbastress: The delegate at least needs the `id-kp-OCSPSigning` EKU, right?

The only other potential issue I see is if `BasicResponse.Certs[0]` is not the delegate -- i.e. if it is just an intermediate for validating the signer, for validating a delegate in a later index, or totally unrelated. If the spec mandates that the certificate, when present, must be the delegate, then a comment calling that out would be helpful. If not, I don't see a reasonable way to pick one short of scanning all of `Certs` for candidates and then checking the signature.

casebenton: Oh good catch, the delegate definitely needs proper EKU designation. 

From the RFC: "This [delegate] certificate MUST be issued directly to the responder by the cognizant CA" (https://tools.ietf.org/html/rfc6960#section-2.6). I'm interpreting that as "no intermediates allowed," do you think that's correct? Or could that just mean that the delegate chain must link back to the issuing CA?

x509/revocation/crl/crl.go
casebenton: @dadrian how does https://github.com/zmap/zcrypto/pull/142/commits/2715be69b03d0f884f021006341686b441d7aa31 look?

x509/revocation/crl/crl.go
```diff
@@ -0,0 +1,154 @@
...
+
+// ListExtensionData - Data from optional, non-critical pkix.CertificateList extensions
+type ListExtensionData struct {
+	CRLNumber int
+	AuthKeyID x509.SubjAuthKeyId `json:"authority_key_id,omitempty"`
+}
+
+// RevokedCertExtensionData - Data from optional, non-critical pkix.RevokedCertificate extensions
+type RevokedCertExtensionData struct {
+	Reason         string `json:"revocation_reason,omitempty"`
```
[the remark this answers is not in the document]
casebenton: Oh good call, thanks.

x509/revocation/crl/crl.go
```diff
@@ -12,46 +12,49 @@ import (
...
-	case PrivilegeWithdrawn:
-		return "privilegeWithdrawn"
-	case AACompromise:
-		return "aACompromise"
-	default:
-		return "Unknown revocation reason code: " + strconv.Itoa(int(r))
+// UnmarshalJSON implements the json.Unmarshaler interface
+func (code *RevocationReasonCode) UnmarshalJSON(b []byte) error {
+	aux := struct {
+		Code  int
```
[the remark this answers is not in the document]
casebenton: Should I move my definitions to https://github.com/zmap/zcrypto/blob/master/zcrypto_schemas/zcrypto.py? Or was that just an example of how types are defined?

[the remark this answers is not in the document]
justinbastress: The way it works today, you would update the schema.py to match the JSON emitted by your go code -- Alex is working on a way to reduce this duplication, but I don't think it's been finalized yet.

Currently the schema updates are only needed if your results are going into BigTable / ElasticSearch, since that's what we use to generate the schema.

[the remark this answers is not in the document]
casebenton: Ok, I don't think that these results are going into BigTable just yet. Maybe I can plan on updating schema.py when it's time to start adding these results to BigTable?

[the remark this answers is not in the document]
justinbastress: Yeah, I don't see any reason the schema needs to be updated immediately.

casebenton:
@dadrian @justinbastress I think that I've gone through all the comments here so far.

justinbastress:
The only correctness issue I still see is the corner case where `len(response.Certs) > 0` but `response.Certs[0]` is not the delegate signer (for instance, if it's just a certificate to be used for verifying the non-delegate issuer), at https://github.com/zmap/zcrypto/pull/142/files#diff-b1c1d783dc52fb97676b6932493fa2afR305.

casebenton:
Ok, that should be resolved in https://github.com/zmap/zcrypto/pull/142/commits/b752362be24287d8c4f742ab63f3adcdaafc7680

casebenton:
Right, good call, I just fixed both of those issues.

PR #145, opened 2018-08-03
Removes validation_policy=warn override from OID fields.

clayzim:
Lets errors be logged accurately for invalid OID fields.


justinbastress:
Isn't this going to cause problems? I thought this was added because there were too many places (perhaps some out of our control?) where optional OIDs were not `omitempty`'d.

clayzim:
I'm not sure. My thought was to let them error out so that I could assess the issue from examples in the dead-letter queue. If there's an established reason for this override, that's fine. What other method should I use to find the source of this error?

PR #148, opened 2018-08-06
Add Browser Revocation Support: Mozilla, Google, Microsoft

casebenton:
New libraries for checking revocation status of certificates through Mozilla OneCRL, Google CRLSet, and Microsoft DisallowedCert. 

Testing Coverage:
Mozilla 86%
Microsoft 87%
Google 68% 
(remaining uncovered code is mostly error handling for malformed raw revocation lists)


x509/revocation/google/google.go
```diff
@@ -0,0 +1,381 @@
...
+		return nil, err
+	}
+	return Parse(crlSetReader, version)
+}
+
+// Check - Given a parsed CRLSet, check if a given cert is present
+func (crlSet *CRLSet) Check(cert *x509.Certificate, issuerSPKIHash string) *Entry {
+	issuersRevokedCerts := crlSet.IssuerLists[issuerSPKIHash]
+	if issuersRevokedCerts == nil { // no entries for this issuer
+		return nil
```
justinbastress: Would it be useful to distinguish found-but-not-revoked from no-CRL-for-this-issuer?

casebenton: I can't think of a reason that it would be necessary. Returning at line 58 effectively means that "there are other certs signed by the same parent that are now revoked," but once we run the scanner on all certs it would be straightforward to determine this in other ways. (Line 57 being true doesn't imply that there's no revocation data available, just that there are no certs revoked by CRLSet with this issuer.) Maybe there's some situation I haven't considered though?

PR #144, opened 2018-07-27
changes to support *25519 DH and Signing keys

dopey:
ed25519 is included in TLS 1.3. Hopefully a good enough reason to start including it.

PR #155, opened 2018-08-17
Clean up and remove old x509 code, condense files, investigate 1.10 migration

### LINKED ISSUE #1: Backport changes from golang 1.8 x509 lib

casebenton:
Version 1.10  of the standard `crypto/x509` library introduces features which zcrypto already supports independently, for example SANExtension parsing in certificates. Resolving these differences would risk breaking dependencies, with little clear benefit, so it seems that diverging from the standard `crypto/x509` and pkix libraries is the most reasonable approach moving forward.

This PR condenses the `zcrypto/x509` package to remove unused and old code. This includes code for handling root stores, which should now be done with `zmap/rootfetch` or with other means. It also removes the *_updated.go paradigm, as it does not serve a clear purpose if we adopt a policy of
longer strictly follow updates to the standard library.

This and #108 effectively close #1.


x509/README.md
```diff
@@ -1,6 +1,8 @@
...
+this package has now diverged enough that it is no longer
+updated with direct correspondence to new go releases.
 
-Modifications to shared files are located in $name_modified.go to 
-facilitate tracking the stdlib updates. 
\ No newline at end of file
+Approximately supports all the features of
+github.com/golang/go/crypto/x509 package at:
+branch: release-branch.go1.10
+revision: dea961ebd9f871b39b3bdaab32f952037f28cd71
```
dadrian: Do we actually have cipher parity with Golang stdlib, or are we missing the 25519 ciphers (and maybe the new numbers for the Salsa and ChaCha ciphers)?

casebenton: Yes, it looks like we have cipher parity with the Golang stdlib for `/x509`. We cover every signature and public key OID that is covered by them. Interestingly, the golang stdlib `/x509` does not yet support `*25519`, even though it is supported in the stdlib's `/tls`. When #144 lands we will be ahead of them in `*25519` support.

casebenton: We may not be at cipher parity for `/tls` (probably for `*25519`), but we may as well wait until the stdlib supports 1.3 (which could be a while, it's not in the upcoming golang v1.11) before bringing our `/tls` up to date.

dadrian: Good point, I was conflating x509 and tls.

PR #162, opened 2018-08-22
Regenerate x509/extended_key_usage.go to add support for 1.3.6.1.5.5.8.2.2

### LINKED ISSUE #95: Add 1.3.6.1.5.5.8.2.2
We should detect when 1.3.6.1.5.5.8.2.2 extended key usage is present.

Here's an example certificate:

https://crt.sh/?id=209727518&opt=cablint

casebenton:
It looks like running `go generate` on my machine spit out the generated file with lots of constants listed in a different order, so the diff is highly inflated. Only change should be addition of `ExtKeyUsageIpsecIntermediateSystemUsage` and related methods and definitions. 

Closes #95.

### DIFF
--- a/x509/extended_key_usage.go
+++ b/x509/extended_key_usage.go
@@ -71,6 +71,7 @@ OID_EKU_IPSEC_TUNNEL = "1.3.6.1.5.5.7.3.6"
 OID_EKU_IPSEC_USER = "1.3.6.1.5.5.7.3.7"
 OID_EKU_TIME_STAMPING = "1.3.6.1.5.5.7.3.8"
 OID_EKU_OCSP_SIGNING = "1.3.6.1.5.5.7.3.9"
+OID_EKU_IPSEC_INTERMEDIATE_SYSTEM_USAGE = "1.3.6.1.5.5.8.2.2"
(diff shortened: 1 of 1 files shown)


casebenton:
@zakird the only weird thing is that the `oid` tags are now missing, compare https://github.com/zmap/zcrypto/pull/162/files#diff-9fc381c98056b088c2aeda35cfe9651bL267 to https://github.com/zmap/zcrypto/pull/162/files#diff-9fc381c98056b088c2aeda35cfe9651bR249

casebenton:
It looks like those `oid` tags were added in https://github.com/zmap/zcrypto/pull/122/files#diff-9fc381c98056b088c2aeda35cfe9651bR206. I'm having trouble finding the auto-generate code that would have created this, @justinbastress how did you add this in?

justinbastress:
I’m not sure; it seems like there was a python script, but I don’t know where it would be checked in.

casebenton:
The non-determinism resolved by #163 is what caused the massive changes in the file. Missing OID labels are also resolved by #163. After rebase, the diff is much more reasonable.

casebenton:
@zakird r?

PR #163, opened 2018-08-23
Fix non-deterministic eku generation

casebenton:
Previously, the generation script for `extended_key_usage.go` used range-based for loops on maps. Golang does not order map keys, so the generation of extended_key_usage.go was non-deterministic. This PR changes the generation script so that the output is deterministic. I also re-generated `extended_key_usage.go` with the new script so that future re-generation PRs are greatly simplified. 

Also ensures that OID tags are included as shown in https://github.com/zmap/zcrypto/pull/122, specifically at https://github.com/zmap/zcrypto/pull/122/commits/22e806f59648401f7825f5fe9841aa01fd7ee84b#diff-9fc381c98056b088c2aeda35cfe9651bR206.

### DIFF
--- a/x509/extended_key_usage_gen.go
+++ b/x509/extended_key_usage_gen.go
@@ -117,7 +117,7 @@ import (
 func generateASN1(oidToName map[string]OID) []byte {
 	buffer := bytes.Buffer{}
 	// Create sorted slice of keys to ensure deterministic output
-	var keys []string
+	var keys = make([]string, 0, len(oidToName))
 	for k := range oidToName {
 		keys = append(keys, k)
 	}
@@ -139,7 +139,7 @@ func generateIntegerConstants(oidToName map[string]OID) []byte {
 	buffer.WriteString("const (\n")
 	first := true
 	// Create sorted slice of keys to ensure deterministic output
-	var keys []string
+	var keys = make([]string, 0, len(oidToName))
 	for k := range oidToName {
 		keys = append(keys, k)
 	}
@@ -161,7 +161,7 @@ func generateIntegerConstants(oidToName map[string]OID) []byte {
 func generateNameConstants(oidToName map[string]OID) []byte {
 	buffer := bytes.Buffer{}
 	// Create sorted slice of keys to ensure deterministic output
-	var keys []string
+	var keys = make([]string, 0, len(oidToName))
 	for k := range oidToName {
 		keys = append(keys, k)
 	}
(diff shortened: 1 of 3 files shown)


x509/extended_key_usage_gen.go
```diff
@@ -145,7 +160,14 @@ func generateIntegerConstants(oidToName map[string]OID) []byte {
 
 func generateNameConstants(oidToName map[string]OID) []byte {
 	buffer := bytes.Buffer{}
-	for _, oid := range oidToName {
+	// Create sorted slice of keys to ensure deterministic output
+	var keys []string
```
zakird: This is probably better off as an array than a slice so that it doesn't keep getting expanded.

x509/extended_key_usage_gen.go
```diff
@@ -145,7 +160,14 @@ func generateIntegerConstants(oidToName map[string]OID) []byte {
...
+	for k := range oidToName {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		oid := oidToName[k]
```
zakird: Seems like an name should come out of `oidToName` not an OID.

x509/extended_key_usage_gen.go
```diff
@@ -159,7 +181,15 @@ func generateOIDMap(oidToName map[string]OID, mapName string) []byte {
...
-	for _, oid := range oidToName {
+
+	// Create sorted slice of keys to ensure deterministic output
+	var keys []string
+	for k := range oidToName {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		oid := oidToName[k]
```
zakird: Same comments as above.

x509/extended_key_usage_gen.go
casebenton: This map has a confusing name, it looks like it maps raw OIDs (represented as strings) straight from the `zmap/constants` csv into the `OID` type. https://github.com/zmap/zcrypto/blob/master/x509/extended_key_usage_gen.go#L253

So it's more like `rawToOID` or something like that. I think that I'll rename the maps for clarity instead.

x509/extended_key_usage_gen.go
casebenton: The size of `oidToName` isn't known at compile time, so I don't think I could use an array - what about slice with an initial capacity big enough so that no slice expansion is required? Something like `var keys = make([]string, 0, len(oidToName))`

x509/extended_key_usage_gen.go
casebenton: (also `oid := oidToName[k]` was effectively what was happening before my PR when the range-based for loop was used directly on oidToName)

zakird:
All sounds good.

PR #172, opened 2018-08-29
Add unmarshalers for certificates

### LINKED ISSUE #169: Add Unmarshalers to x509/json
Sub-issue of #78. 

There are several types in `x509/json`, most notably `Certificate`, which have JSON marshalers but no unmarshalers. 

From #78: 
> We should ensure we have Unmarshalers for anything that has a Marshaler.

Tests are also needed for any new unmarshaler methods added.

casebenton:
The `MarshalJSON` method on `x509.Certificate` condenses data in a way that is not recoverable. Given this situation, we can only fully reconstruct an `x509.Certificate` if we have the raw asn1 data. Therefore, the UnmarshalJSON methods on `x509.Certificate` and `x509.jsonCertificate` have been set to always error. A new type, `JSONCertificateWithRaw`, is introduced for parsing json that includes the raw asn1 cert data, from which a `x509.Certificate` can be constructed using the `x509.ParseCertificate` function.  

Closes #169, since all other Marshalers in `x509/json` have corresponding Unmarshalers.

### DIFF
--- a/x509/json.go
+++ b/x509/json.go
@@ -10,6 +10,7 @@ import (
 	"crypto/rsa"
 	"encoding/asn1"
 	"encoding/json"
+	"errors"
 	"net"
 	"sort"
 
@@ -270,7 +271,6 @@ type JSONSubjectKeyInfo struct {
 	SPKIFingerprint CertificateFingerprint `json:"fingerprint_sha256"`
 }
 
-
 // JSONSignature - used to condense several fields from x509.Certificate
 // related to the signature into one field within JSONCertificate
 // Unfortunately, this struct cannot have its own Marshal method since it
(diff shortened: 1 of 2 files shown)


x509/json.go
```diff
@@ -382,6 +382,42 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
...
+}
+
+// JSONCertificateWithRaw - intermediate struct for unmarshaling json
+// of a certificate - the raw is require since the
+// MarshalJSON method on Certificate condenses data in a way that
+// makes extraction to the original in Unmarshal impossible.
+// The JSON output of Marshal is not even used to construct
+// a certificate, all we need is raw
+type JSONCertificateWithRaw struct {
+	Raw []byte
```
dadrian: This needs `json:"raw,omitempty"` tag

PR #175, opened 2018-08-30
Mark self-signed certs as valid signatures for json ouput

### LINKED ISSUE #72: Self-signed certificates always have signature.valid = false
Self-signed certificates should only have signature.valid = false when they're self-signed (i.e. issuer matches subject), but the signature is actually not from the key in the certificate.

casebenton:
Closes #72


x509/json_test.go
```diff
@@ -21,15 +21,15 @@ func TestCertificateJSON(t *testing.T) {
...
...com"],"ip_addresses":["1.2.3.4"],"other_names":[{"id":"1.2.3.4","value":"DCBEQlZ6YjIxbElHOTBhR1Z5SUdsa1pXNTBhV1pwWlhJPQ=="}],"registered_ids":["1.2.3.4"],"uniform_resource_identifiers":["http://www.insecure.com"]}},"signature":{"signature_algorithm":{"name":"SHA1-RSA","oid":"1.3.14.3.2.29"},"value":"NKUyMbXEPqjfHmhThDqPF5onrwqVAtumI0FQtfZs7V1Ve4fPU+Hc7jNvyGzv3h2qynKORbX+ZgXX1yUaVL3VJBpKogwr8ogpVbr5q/D5KrIwI8nUHF9meMPjDj0imtuE4KqSSvHa9YqT16e1iMwcs9vYEmpetIVs8pGIUp9DtpSKV9sn0ZlaLRO4lA1Cx25N8W5Ue3Qoj78Vl/lZCIBn4Whqpo/h5joOlETNyghzm8Cw0PAMviC/k/rFstizrWa1c8s60Ex5S3A4cdIvF03U5mwYSGMxgfi7TvD6uRlp/L7t6VSxHChq3C9j965Xec2qK6klS/4XXVAizN7/V/6TEA==","valid":true,"self_signed":true},"fingerprint_md5":"b930cccf8bf02db782dadb8e7171d783","fingerprint_sha1":"51e9e3ddf2d7d46bd269851efbcb574aa4ac47c1","fingerprint_sha256":"2b96e909ccec2ef95d0ef20678a05844c46308d44a78958b4fa9e474a522683b","tbs_noct_fingerprint":"7729889cecc4fd392a4b1bff3222093345e3cc0a7b8f35b1fb3b2a7201818093","spki_subject_fingerprint":"d417ddc01fb3d88a50ed9dca6fd7e1484e6043dc64fc743fbb17f57af10cc13b","tbs_fingerprint":"7729889cecc4fd392a4b1bff3222093345e3cc0a7b8f35b1fb3b2a7201818093","validation_level":"unknown","redacted":false}`,
```
casebenton: I changed these two tests to expect that `signature.valid: true`, since they're both signed (previous test expected them to be marked as not valid signatures)

PR #179, opened 2018-09-11
changes to support *25519 DH and Signing keys + Conflict Resolution

zakird:
This is @dopey's PR along with some changes such that it can be merged in master. From @dopey's original PR:

> ed25519 is included in TLS 1.3. Hopefully a good enough reason to start including it.


x509/x509.go
```diff
@@ -200,6 +210,7 @@ const (
 	SHA256WithRSAPSS
 	SHA384WithRSAPSS
 	SHA512WithRSAPSS
+	ED25519SIG
```
justinbastress: That's not the right const name is it? Though, a quick search isn't turning up any -- so maybe there isn't a "standard" name here, in which case `ED25519Sig` would match the camel case of the rest (the `Sig` suffix itself is inconsistent, but I don't know how else you would distinguish it from just `ED25519`)

x509/x509.go
```diff
@@ -352,6 +371,7 @@ var signatureAlgorithmDetails = []struct {
 	{ECDSAWithSHA256, oidSignatureECDSAWithSHA256, ECDSA, crypto.SHA256},
 	{ECDSAWithSHA384, oidSignatureECDSAWithSHA384, ECDSA, crypto.SHA384},
 	{ECDSAWithSHA512, oidSignatureECDSAWithSHA512, ECDSA, crypto.SHA512},
+	{ED25519SIG, oidKeyEd25519, ED25519, crypto.Hash(0)},
```
justinbastress: It would be nice to have something like `crypto.NoDigest` instead of manually creating `crypto.Hash(0)`...this is consistent with the existing `MD2WithRSA` case, but the reasoning seems to be different: presumably the MD2 case is using crypto.Hash(0) because MD2 just isn't supported, rather than there actually being no hash.

x509/x509.go
```diff
@@ -998,17 +1030,16 @@ func CheckSignatureFromKey(publicKey interface{}, algo SignatureAlgorithm, signe
 	//case MD2WithRSA, MD5WithRSA:
 	case MD2WithRSA:
 		return InsecureAlgorithmError(algo)
```
justinbastress: As an aside, zcrypto should probably allow this to proceed, no?

zakird:
@dopey: wanted to tag you specifically since I'm not sure if you get notifications about this PR in case you have any comments for @justinbastress.

x509/x509.go
```diff
@@ -2321,8 +2380,15 @@ func signingParamsForPublicKey(pub interface{}, requestedSigAlgo SignatureAlgori
 			err = errors.New("x509: unknown elliptic curve")
 		}
 
+	case ed25519.PublicKey:
+		// ed25519.PublicKey doesn't return a pointer to the key so we check for both
+		pubType = ED25519
+		hashFunc = 0
+		shouldHash = false
+		sigAlgo.Algorithm = oidKeyEd25519
```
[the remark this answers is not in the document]
dopey: I'm down to err on the side of consistency here. Looks like Ed25519 is the most consistent name for the algorithm, so changing `ED25519` -> `Ed25519` makes sense to me.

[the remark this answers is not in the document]
dopey: Making this change now. Will push in a moment.

x509/x509.go
dopey: I didn't realize the `cryptyo.Hash(0)` in MD2 was do to lack of support, so I definitely agree that clarifying the reasoning would be useful. I can add a `cryptoNoDigest` var above (with some documentation explaining that it should be used if/when there is no hash).

dopey:
I pulled this branch and pushed up a few changes to address the comments. https://github.com/zmap/zcrypto/compare/smallstep-25519...smallstep:smallstep-25519

zakird:
I just merged those commits into this branch. This looks fine to me.

dopey:
whoops. I'm a dope and forgot to update the tests. just pushed again. sorry 😬

zakird:
I'm not 100% sure we never merged this, I think it was because tests weren't passing at the time. It looks fine though, and I'm going to take a quick look and merge.

PR #186, opened 2019-03-07
Update CI to 1.11.x, add Go module support,

cpu:
:wave: Hi everyone,

I made the assumption that like for `zlint` there's not a desire to have multiple Go versions in CI, just a currently supported release stream. I chose Go 1.11.x for this because it has module support and that's what `zlint` has been using since https://github.com/zmap/zlint/commit/bb32118ad3ab29c4d9a697aa1d8faa71c07e7500 There is a Go 1.12.x release available now but I haven't had time to kick its tires yet. If you'd prefer to jump right to Go 1.12.x I'll give it a go.

One potential follow-up [I flagged](https://github.com/zmap/zlint/pull/266#discussion_r263547237) in https://github.com/zmap/zlint/pull/266 is that the `github.com/asaskevich/govalidator` dep. could be removed from both `zcrypto` and `zlint` if `zcrypto` provided its own `IsURL` utility function: https://github.com/zmap/zcrypto/issues/187


cpu:
> cpu force-pushed the cpu:cpu-go-mod  branch from 4183cc6 to 1365dcb 6 minutes ago

nothing to see here

Issue #188, opened 2019-03-14
Parse TorServiceDescriptorHash x509 extension


cpu:
Based on [feedback](https://github.com/zmap/zlint/pull/263#discussion_r261844546) on a `zlint` PR (https://github.com/zmap/zlint/pull/263) `zcrypto` is the correct place to parse the `TorServiceDescriptorHash` x509 extension (oid `2.23.140.1.31`) described initially in [CABF Ballot 144](https://cabforum.org/2015/02/18/ballot-144-validation-rules-dot-onion-names/).

I'm planning to submit a PR for this in the next week or so to complete https://github.com/zmap/zlint/issues/14 

I need an issue to reference in my TODO list so I don't forget about it :-)

PR #189, opened 2019-03-16
x509: support parsing CABF TorServiceDescriptor extension.

cpu:
In CAB Forum [Ballot 144][144] validation rules for `.onion` names were added to the EV Guidelines. The rules were subsequently tweaked in [Ballot 198][198] (declared invalid) and [Ballot 201][201]. The rules (See Appendix F of the [EV Guidelines](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-EV-Guidelines-v1.6.8.pdf)) specify that certificates with an `.onion` subject must include a `TorServiceDescriptor` x509 extension with OID `2.23.140.1.31`. This extension is an ASN.1 `SEQUENCE` of `TorServiceDescriptorHash` entries.

This commit adds support for recognizing and parsing the `TorServiceDescriptor` extension when parsing certificates. The individual `TorServiceDescriptorHash` objects are made accessible on the parsed `x509.Certificate` as the `TorServiceDescriptors` field, a slice of `x509.TorServiceDescriptorHash` objects:
```
type TorServiceDescriptorHash struct {
	Onion    string  // The Onion URI (e.g. "https://zmap.onion")
	AlgorithmName      string  // A friendly name for the hash algorithm in use (e.g. "SHA256")
        Algorithm  pkix.AlgorithmIdentifier // The raw pkix algorithm identifier.
	Hash     CertificateFingerprint  // Hash of the DER encoded ASN.1 of the onion service's public key
	HashBits int     // Number of bits in Hash. (e.g. 256)
}
```

The JSON marshaling of `x509.Certificate` objects is updated to marshal this new field as `tor_service_descriptors`.

Example `zcertificate` output (using `zcrypto` from this branch) against [1231238276.crt][crt], an example DigiCert issued certificate found in CT: 

```
$ zcertificate 1231238276.crt | jq .parsed.extensions.tor_service_descriptors
INFO[0000] reading from 1231238276.crt                  
INFO[0000] writing to stdout                            
[
  {
    "onion": "https://hzwjmjimhr7bdmfv2doll4upibt5ojjmpo3pbp5ctwcg37n3hyk7qzid.onion",
    "algorithm_name": "SHA256",
    "hash": "d6bae93a21d29770f49e2003eb7a2cda08e9e7c3b4281adee8bd5edd1bd3e87b",
    "hash_bits": 256
  }
]
```

Resolves https://github.com/zmap/zcrypto/issues/188

[144]: https://cabforum.org/2015/02/18/ballot-144-validation-rules-dot-onion-names/
[198]: https://cabforum.org/2017/05/08/ballot-198-onion-revisions/
[201]: https://cabforum.org/2017/06/08/2427/
[crt]: https://crt.sh/?id=1231238276


cpu:
I started working on the updated `zlint` side of this and realized I was being too aggressive in generating errors in the `zcrypto` parsing. I dialed things back in 0577d9d - now basically just structural parsing errors and trailing data errors are passed back through `x509.ParseCertificate`. Invalid UTF8 bytes, unknown hash algorithms, mismatched algorithm vs hash bytes, etc will not cause errors as of 0577d9d. I'll enforce those properties in the `zlint` linter.

cpu:
> Actually, this naming scheme seems to mismatch how we've named other fields. Other fields are all lowercase, underscores instead of camel case. We should probably match for this.

@zakird I'm not sure I follow, the struct field name and JSON field name seem to match the other `CertificateExtensions` in `x509/extensions.go`:

https://github.com/zmap/zcrypto/blob/d3043756f9f43af2ff0ecd05ff83e3e2abc26437/x509/extensions.go#L36-L50

Can you give me more specific feedback about which name you want changed? Thanks!

zakird:
Sorry about that! The name of the extension itself looks fine: `tor_service_descriptors` is in line with other extensions. My concern is over the naming of the inner elements: https://github.com/zmap/zcrypto/pull/189/files#diff-e1a61098d80f22c20a23992446fce278R1200

cpu:
@zakird Ahhhh! Now we're on the same page :-) I'll add `json` tags with correct snake case names to the `TorServiceDescriptorHash` struct fields in `x509.go` :+1:

cpu:
@zakird I think 6507711 accomplishes what you want. I also updated the `zcertificate` output example in the PR desc.

Along with the rename of the fields I also removed the `pkix.AlgorithmIdentifier` field (`Algorithm`) from the JSON output. I don't think its high level enough to merit being put into the marshalled JSON without making a nicer representation.

x509/x509.go
```diff
@@ -1183,6 +1188,23 @@ type distributionPointName struct {
...
+// Each TorServiceDescriptorHash holds an onion URI (a utf8 string with the
+// .onion address that was validated), a hash algorithm name (computed based on
+// the pkix.AlgorithmIdentifier in the TorServiceDescriptorHash), the hash bytes
+// (computed over the DER encoding of the ASN.1 SubjectPublicKey of the .onion
+// service), and the number of bits in the hash bytes.
+type TorServiceDescriptorHash struct {
+	Onion         string                   `json:"onion"`
+	Algorithm     pkix.AlgorithmIdentifier `json:"-"`
+	AlgorithmName string                   `json:"algorithm_name"`
+	Hash          []byte                   `json:"hash"`
```
dadrian: We usually output hashes as hex, not Base64. There's an alias type somewhere in ZCrypto (maybe called `CertificateFingerprint`?) that does this transparently. I'd update the JSON serialization and/or type definition of `TorServiceDescriptorHash` to use a `Fingerprint` type.

x509/testdata/onionSANBadServDescAlgorithm.pem
```diff
@@ -0,0 +1,12 @@
+-----BEGIN CERTIFICATE-----
```
dadrian: I'm not going to make you provide a programmatic method for generating these test certificates, but if you do have any scripting or even a link to some certs on Censys, just drop it in this directory.

x509/x509.go
cpu: Sounds good, can do.

x509/testdata/onionSANBadServDescAlgorithm.pem
cpu: I'm afraid I don't :-( I made these certs by hand tweaking a cert template in a one-off Go program. I didn't save intermediate versions of the program as I went along, just the rendered PEM encoded certificates. In retrospect that was pretty hackish.

I think rather than provide scripting/links to test certs it would be cleaner overall to re-implement the tests as in-memory `pkix.Extension` instances handed directly to `parseTorServiceDescriptorSyntax` instead of PEM encoded certificates given to `ParseCertificate`. I mostly did the latter because that's how I wrote the `zlint` tests where it made more sense to operate at the certificate level. If I do the former I think the tests will be self-documenting and won't rely on opaque test data.

Would that work for you?

cpu: I implemented the above in 9458ee5 and deleted the `.pem` files. The testcase code I wrote is now pretty much identical to the relevant parts of the throw away go programs I wrote to generate the `.pem`. I think this is the better approach, thanks for the nudge! :+1:

x509/x509.go
cpu: Done as of a7bcfac :ballot_box_with_check:

cpu:
@dadrian I think this is ready for another :mag: when you have a chance. Thanks!

cpu:
@dadrian If you're able to give this another :mag: today I'll have time to handle any revisions required on Saturday morning. If you can't that's OK too, I'll pick it up again next week.

PR #190, opened 2019-03-16
 implement unmarshal interface for SignatureAndHash

kwang40:
I implement the unmarshal interface for SignatureAndHash. For reverse lookup from name to hash/signature, I just iterate through the maps. Since the numbers of items in the maps are super small, it should not affect the performance.


dadrian:
@justinbastress please merge this if it looks good to you

justinbastress:
Looks OK to me. IIRC there were some issues mapping back and forth, but I don't see any many-to-one cases in the maps here.

PR #194, opened 2019-07-11
ci: enforce all files are gofmt'd.

cpu:
Additionally run all existing files through `gofmt -s -w` to ensure CI will pass with the new test.


cpu:
> The Travis CI build failed
Blech. The `go fmt` in CI isn't respecting the vendor deps and generating output. Will iterate in a moment.

cpu:
Almost there. CI is passing now but seems to have run two jobs in a matrix where I wanted one job with both env vars set.

cpu:
Getting this to work with the vendored dependencies is trickier than expected... [This issue](https://github.com/golang/go/issues/27841) indicates it _should_ work. I thought my trouble was due to process substitution being whacky but I know the `GOFLAGS` I added directly to the `go fmt` are being processed because my typo in b44c9c9 broke the `go fmt`, but the correct value in ef2668c isn't being respected so I'm a bit stumped :thinking:

I'm going to try cloning outside of my GOPATH and working on a fix locally instead of spamming the CI queue with my flailing :sweat: I'll reopen this PR shortly.

cpu:
Re-opened in a working state in https://github.com/zmap/zcrypto/pull/195

In a bit of cosmic irony we had already bumped into this exact problem in Boulder and arrived at the work-around I used in 195:
https://github.com/letsencrypt/boulder/blob/74699486ec3338c99a633459fa553c586366f6ca/test.sh#L72-L74

PR #195, opened 2019-07-11
ci: use GOFLAGS, enforce gofmt -s.

cpu:
This PR sets up a global `GOFLAGS` to avoid needing to repeat it for each command.

It also changes the `script` to exit on the first error, and to enforce that all non-vendored `.go` files satisfy `gofmt -s` without diffs. This will make sure all code is consistently formatted and help contributors
do the right thing by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,24 @@ go:
   - "1.11.x"
 
 env:
-  - GO111MODULE=on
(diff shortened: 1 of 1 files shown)


cpu:
Still not quite right :sweat: Forgot to expect silence.

Issue #198, opened 2019-07-26
Populate a map of OID -> pkix.Extension in Certificate objects.


cpu:
**tl;dr - I think `zcrypto` should offer a `map[string]pkix.Extension` field on `x509.Certificate` to avoid lots of linear searching in `zlint`'s many lints.**

In [zlint](https://gitub.com/zmap/zlint) a common pattern in the `CheckApplies(c *x509.Certificate) bool` functions of individual lints is to check a `zcrypto.X509.Certificate` for a specific x509 extension. e.g. consider the [`lint_ext_san_empty_name`](https://github.com/zmap/zlint/blob/d9a29c3ddfb5b3b498dea53607b9c395267d3807/lints/lint_ext_san_empty_name.go#L42-L44) lint's `CheckApplies` function:

```
return util.IsExtInCert(c, util.SubjectAlternateNameOID)
```

The `util.IsExtInCert` function in turn relies on a `util.GetExtFromCert` function that [performs a linear search](https://github.com/zmap/zlint/blob/d9a29c3ddfb5b3b498dea53607b9c395267d3807/util/oid.go#L110-L119) over the `zcrypto.X509.Certificate`'s `Extensions` field:
```
	for i := range cert.Extensions {
		if oid.Equal(cert.Extensions[i].Id) {
			return &(cert.Extensions[i])
		}
	}
```

Looking across the whole `zlint` codebase with a quick `ack` I see there are **200** references to `util.GetExtFromCert` or `util.IsExtInCert`:

```
$ ack -c -h "GetExtFromCert|IsExtInCert" $GOPATH/src/github.com/zmap/zlint
200
```

Benchmarking confirms that ` github.com/zmap/zlint/util.GetExtFromCert` and `encoding/asn1.ObjectIdentifier.Equal` are present int the top 10 nodes accounting for cpu time in a `pprof` trace:

```
$ cd $GOPATH/src/github.com/zmap/zlint
$ go test --run=XXX -bench=. -cpuprofile all.profile
<snipped>
$ go tool pprof all.profile
zlint.test
Type: cpu
Time: Jul 26, 2019 at 10:58am (EDT)
Duration: 4.13mins, Total samples = 4.79mins (115.90%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top10
Showing nodes accounting for 146.71s, 51.07% of 287.25s total
Dropped 522 nodes (cum <= 1.44s)
Showing top 10 nodes out of 221
[...]

PR #200, opened 2019-07-29
x509: add ExtensionsMap to Certificate.

cpu:
A new `ExtensionsMap` field is added on the `x509.Certificate` object. This field maps from the string form of a `pkix.Extension.Id` OID (e.g. "1.3.6.1.5.5.7.1.1") to a raw `pkix.Extension` value. This lets relying code check if a certificate has a given extension by OID in O(1) time.

To accomplish this the `parseCertificate` function is updated to populate this map at the same time as parsing the existing `x509.Certificate.Extensions` field. The `x509.JSONCertificate` type and `x509.Certificate.MarshalJSON` function are **not** updated to avoid duplicate and unnecessary extension output in JSON.

Resolves https://github.com/zmap/zcrypto/issues/198


cpu:
I'll start on the `zlint` update PR once this is all merged up and I can bump the `go.mod` ref.

zakird:
Hrm, actually, I think we probably don't want to output these in the JSON. That would lead to every extension showing up twice in the output.

cpu:
@zakird :+1: I'll back out the JSON updates and fix the PR description.

cpu:
>  I think we probably don't want to output these in the JSON.

Fixed in 9e2dce4

zakird:
🙌

zakird:
Is there the potential to have certificates that have multiple extensions with the same OID? How do we want to handle? It seems like this might be a construction issue and the type of thing we'd want to handle in ZLint.

cpu:
> How do we want to handle? It seems like this might be a construction issue and the type of thing we'd want to handle in ZLint.

I think ignoring the possibility for the purpose of the `x509.Certificate.ExtensionsMap` and then adding a lint on the `Zlint` side that iterates `x509.Certificate.Extensions` and checks for duplicates would be my preference. I can't think of any adverse side effects from this approach but I'm open to input if I missed something. Any duplicate OIDs squashed to one entry in the `x509.Certificate.ExtensionsMap` would be present in duplicate in the original `x509.Certificate.Extensions` slice.

zakird:
That's fine with me. The alternative would be to map str -> []ext but if that's _never_ the expected behavior, then that seems burdensome to API users. I'll merge this and we can handle that in ZLint if that sounds good.

cpu:
I added a small clarification note to the `ExtensionMap` Godoc comment since I think its worth capturing this accepted limitation in the doc string: 105ca31

cpu:
>  The alternative would be to map str -> []ext but if that's never the expected behavior, then that seems burdensome to API users.

To close the loop here [RFC 5280 Sec 4.2](https://tools.ietf.org/html/rfc5280#section-4.2) confirms this expectation is reasonable and something we can lint for in `zlint` at the error level:

>  A certificate MUST NOT include more than one instance of a particular extension.

cpu:
> I think ignoring the possibility for the purpose of the x509.Certificate.ExtensionsMap and then adding a lint on the Zlint side that iterates x509.Certificate.Extensions and checks for duplicates would be my preference.

@zakird Turns out there is already a lint for this :tada:: https://github.com/zmap/zlint/blob/master/lints/lint_ext_duplicate_extension.go It's O(n^2) but I think that's OK for this one lint.

Issue #203, opened 2019-09-16
Google CRLSet fetch broken


cpu:
:wave: It looks like the `TestFetch` unit test from the `x509/revocation/google` package is consistently failing.  I bumped into this with https://github.com/zmap/zcrypto/pull/202 but it reproduces with Go 1.12.x and seems unrelated. 

I'm not familiar with CRLSet implementation details or Chrome internals but it looks to me like the upstream API has changed and the test failure is a true positive. Here's my quick drive-by diagnosis. 

Adding some debug statements to the test shows `fetch` making a GET request to

 ```
http://clients2.google.com/service/update2/crx?x=id%3Dhfnkpimlhhgieaddgfemjhofmfblmnib%26v%3D%26uc
```
and getting back the XML response:
```
<?xml version="1.0" encoding="UTF-8"?><gupdate xmlns="http://www.google.com/update2/response" protocol="2.0" server="prod"><daystart elapsed_days="4641" elapsed_seconds="46697"/><app appid="hfnkpimlhhgieaddgfemjhofmfblmnib" cohort="1:jcl:" cohortname="Auto" status="ok"><updatecheck status="noupdate"/></app></gupdate>
```

The test is expecting `fetch` to iterate `Apps` elements in the response to pull out a `URL` and `Version` attribute from child `UpdateCheck` elements to find a usable `crxURL`. The response above doesn't include this information in the XML so ultimately `fetch` returns a `errors.New("Failed to parse Omaha response")` `err` and the `TestFetch` unit tests fails.

It seems like the code in this package is adopted from [Agl's crlset-tools](https://github.com/agl/crlset-tools) and I notice that he recently updated the library to account for a new API version: https://github.com/agl/crlset-tools/commit/eb2087904f67e44f1806ae27dcde149e908a4c91

The approach taken in the `crlset-tools` package skips signature checking. I'm not familiar enough with this part of `zcrypto` or its intended use-cases to determine if that's an appropriate trade-off for this package to adopt as well. If it is acceptable then backporting the update from `crlset-tools` will probably fix this problem.

cpu:
> If it is acceptable then backporting the update from crlset-tools will probably fix this problem.

If not checking the signature is fine with folks I can volunteer to do the backporting. If it isn't I think it would be better to task someone with more context to do the needed update, or to consider removing the code from `zcrypto` until it can be fixed.

zakird:
[...]

PR #204, opened 2019-11-12
Store multiple CN and SN values

zakird:
This PR stores multiple values from CN and Serial Number if multiple versions are present rather than simply overwriting with the last one. I'd certainly rather just turn those into arrays, but I don't want to break everything for a lint.

### DIFF
--- a/x509/pkix/pkix.go
+++ b/x509/pkix/pkix.go
@@ -54,6 +54,7 @@ type Name struct {
 	StreetAddress, PostalCode, DomainComponent []string
 	EmailAddress                               []string
 	SerialNumber, CommonName                   string
+	SerialNumbers, CommonNames                 []string
 	GivenName, Surname                         []string
 	// EV Components
 	JurisdictionLocality, JurisdictionProvince, JurisdictionCountry []string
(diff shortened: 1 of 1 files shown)


zakird:
FYI, these tests fail on master. Investigating.

cpu:
> FYI, these tests fail on master. Investigating.

@zakird I took a look at the CI task and it's the same issue I flagged in https://github.com/zmap/zcrypto/issues/203:

```
=== RUN   TestFetch
--- FAIL: TestFetch (0.09s)
=== RUN   TestCheck
--- PASS: TestCheck (0.00s)
FAIL
FAIL	github.com/zmap/zcrypto/x509/revocation/google	0.098s
```

x509/pkix/pkix.go
```diff
@@ -54,6 +54,7 @@ type Name struct {
 	StreetAddress, PostalCode, DomainComponent []string
 	EmailAddress                               []string
 	SerialNumber, CommonName                   string
+	SerialNumbers, CommonNames                 []string
```
cpu: Does there need to be an accompanying diff to `x509/json.go`'s `MarshalJSON` function and `JSONCertificate` type to handle these fields?

zakird: I thought about this too, but I don't think that we necessarily want to output them in JSON. All but a handful of messed up certificates are not going to have more than one CN or Serial Number. I primarily wanted to expose so that we could write a lint that checks for this. If we indeed think that we need to output all in the JSON, I think that might be the time to switch CommonName and SerialNumber to be lists themselves instead of a singular value.

cpu: :+1: Makes sense to me. The `JSONCertificate` godoc explicitly says "This struct has a distinct and independent layout from x509.Certificate, mostly for condensing data across repetitive fields ..." and this does seem like repetitive data.

PR #206, opened 2020-01-22
Backport TLS and x509 from Go 1.13

corny:
Go 1.13 brings TLS v1.3 and still supports SSL v3.

Support has been added for:
* Ed25519 signature algorithm

Support has been dropped for:
* DHE key agreement (only RSA and ECDHE available)
* ExtendedRandom
* ExtendedMasterSecret
* SignedCertificateTimestampExt

Data not (yet) exposed:
* preMasterSecret
* key authentication signature
* private and public keys from the key agreement


dadrian:
There's a lot of ZCrypto-specific functionality beyond just exposing a few methods. Offhand, a few things include:
- Export Ciphers
- RSA Key Exchange
- DH ciphers
- Heartbleed

I'd recommend attempting to come up with this list, figure out what functionality needs to be maintained, and then ensure it's all backported on top of Go 1.13 TLS before continuing.

corny:
Thanks for your response. I'm already trying to figure out which functionality has been added by zgrab. A list would be more than helpful.

corny:
@dadrian  I just found your started backport in #192. I also renamed file to have the prefix `zcrypto_` and I used your functions for marshaling and unmarshaling of `SignatureAndHash`.

What do you think of using two TLS forks in parallel? The current TLS library dropped some depreacted features and cipher suites.

zakird:
Unfortunately, this isn't realistic, in terms of both maintainability and a single handshake being able to accommodate a wide range of servers. We need to maintain a single cohesive library.

corny:
Alright. Do you know any existing test suite for testing TLS clients? It would be nice to run a server with an old OpenSSL version that is also vulnerable to heartbleed. And to have everything in the CI.

mzpqnxow:
I'm wondering if this might provide what is necessary to address #212 ? There is an issue in zgrab2 that is dependent upon this (this is support for server initiated renegotiation)

The zgrab2 issue is here: https://github.com/zmap/zgrab2/issues/252

dissoupov:
@corny what is the status of this PR?
I also started working on porting TLS 1.3 from Go 1.16 and found this PR.

Can it be updated to get the latest from 1.16?

corny:
I'm not working on the backport any more since TLS 1.0 has been deprecated by many browsers.

mzpqnxow:
@dissoupov are you actively working on the TLS 1.3 backport? If so, keep this one in mind: https://github.com/zmap/zcrypto/issues/212

:)

I appreciate any time you put into this, it's going to become more of an issue over time as CloudFlare takes over the rest of the Internet. They've been pushing ESNI since ~2018 and more recently they're pushing [ECH](https://blog.cloudflare.com/encrypted-client-hello/). I believe *both* are TLSv1.3 features, not supported on earlier protocol versions. It seems possible that some of the most high-traffic sites will end up advertising TLSv1.3 only eventually- though I'm not sure how long it will take clients to catch up, so maybe there's plenty of time ...

dissoupov:
Yes, we work in the `feature/TLS1.3` branch.
I just managed to compile and run all the tests.
will submit the first PR soon.

dadrian:
It's been a while since I've thought about this, and when I was at Censys, I was constantly reshelving it because it is a lot of work. I'm not sure where the TLS library in ZCrypto is at these days, but when adding in TLS 1.3, I'd recommend doing the following:

- Dropping support for custom encodings of ClientHello's and so on
- Dropping the support for acting as a FREAK server
- Dropping the flag that allows negotiation of an unimplemented cipher
- (Possibly) dropping Heartbleed testing
- Rethinking the JSON output for the handshake to simply / change things we don't like, as well as adapt to new message structures in TLS 1.3

I think it is important to keep the ability to negotiation SSLv3 through TLS 1.3.

Sorry for the delayed response, I'm usually behind on ZMap Github organizations as it's not my full-time focus anymore (it was from 2013-2020). Do what you will with my branch :)

mzpqnxow:
> 
> I think it is important to keep the ability to negotiation SSLv3 through TLS 1.3.
> 

I agree, this is very, very, very important (at least for me!)

Thanks to both of you for working on this, I know it's a huge amount of effort

PR #214, opened 2020-04-08
Parse OrganizationIdentifier in PXIX Names (2.5.4.97)

dadrian:
The number of files that needs to be edited to do this is oddly high, due to how the JSON code was structured when the X509 parsing was weaker.

Alas.


zakird:
can we get a test perchance? There are a bunch of Certs in Censys with the field.

dadrian:
Yeah, I'll add the one I was using locally.

PR #216, opened 2020-04-16
Parse CABF Organization ID

dadrian:
Parse CABF organization ID based on description here: https://cabforum.org/2019/05/21/ballot-sc17-version-7-alternative-registration-numbers-for-ev-certificates/.

Previously, this would show in `unknown_extensions` with OID `2.23.140.3.1`

Example Output:
```
{
[...]

PR #217, opened 2020-04-19
Parse QC Statements

dadrian:
Parse the QC statements extensions and list the OIDs of each statement. For all of the ETSI QC statements with lints in ZLint, further parse out semantic meaning from each statement.

This includes the following statements/OIDS:
```
oidEtsiQcsQcCompliance      = asn1.ObjectIdentifier{0, 4, 0, 1862, 1, 1}
oidEtsiQcsQcLimitValue      = asn1.ObjectIdentifier{0, 4, 0, 1862, 1, 2}
oidEtsiQcsQcRetentionPeriod = asn1.ObjectIdentifier{0, 4, 0, 1862, 1, 3}
oidEtsiQcsQcSSCD            = asn1.ObjectIdentifier{0, 4, 0, 1862, 1, 4}
oidEtsiQcsQcEuPDS           = asn1.ObjectIdentifier{0, 4, 0, 1862, 1, 5}
oidEtsiQcsQcType            = asn1.ObjectIdentifier{0, 4, 0, 1862, 1, 6}
```

Example output for a certificate with all QC Statements, lifted from the ZLint tests:

```
{
  "version": 3,
  "serial_number": "9591198274709138036589902159",
  "signature_algorithm": {
    "name": "SHA256-RSA",
    "oid": "1.2.840.113549.1.1.11"
  },
  "issuer": {
    "common_name": [
      "Development Sub CA"
    ],
    "country": [
      "PK"
    ],
    "organization": [
      "Development"
    ]
  },
  "issuer_dn": "C=PK, O=Development, CN=Development Sub CA",
  "validity": {
    "start": "2019-09-25T09:37:57Z",
    "end": "2029-10-26T09:37:57Z",
    "length": 318297600
  },
  "subject": {
    "common_name": [
      "Muhammad Bilal Ashraf"
    ],
    "serial_number": [
      "578611675"
    ],
    "country": [
      "PK"
    ],
    "organization": [
      "Development"
    ],
    "given_name": [
      "Muhammad Bilal"
    ],
    "surname": [
      "Ashraf"
    ]
  },
  "subject_dn": "C=PK, O=Development, serialNumber=578611675, GN=Muhammad Bilal, SN=Ashraf, CN=Muhammad Bilal Ashraf",
  "subject_key_info": {
    "key_algorithm": {
      "name": "RSA"
    },
    "rsa_public_key": {
      "exponent": 65537,
      "modulus": "2HPVolNvA0qVuwXPcXIdra/W/VOLzx5RyRjAHUC8ssYKPLJ5rNjUcAFrIMRBELVw1B+Sr3/671cDsh12nFkO4KfCpprvG9gpV2yLZPNhpUPqxpbXa6H6VQv4PYNHuf75kPJzyn2dkhw/RGNfiN9b/qM4LSxHzlrOe+gjN7uSaLEctr16v7lu650lbNC29XfG9FuRKeHAB04WoFtge6P3XwvRkHTVvcUjx0XSRD3Fbcvi/SqKZStFLN382Xp/uA7S+kX0eVHHyIgQyEydOq5lI1mTRCsJn0vOXav2YgZiHGvpDMzXPz/j/iAFblLmGeNFRUQDIrwTYUEfdGjBGOAlVw==",
      "length": 2048
    },
    "fingerprint_sha256": "bc898e576ca5b03513505d6433e91b4ada07f86a6ce3412d088db9580b13500e"
  },
  "extensions": {
    "key_usage": {
      "digital_signature": true,
      "content_commitment": true,
      "value": 3
    },
    "basic_constraints": {
      "is_ca": false
    },
    "subject_alt_name": {
      "email_addresses": [
        "bilal.ashraf@gmail.com"
      ]
    },
    "crl_distribution_points": [
      "http://dev.com/ca.crl"
    ],
    "authority_key_id": "30cd83a736aca535ff211b37406eb2cf5f1c032f",
    "subject_key_id": "5e3b7f1de8d5587eee26b3c792e65ac10e5fbb9d",
    "extended_key_usage": {
      "email_protection": true
    },
    "certificate_policies": [
      {
        "id": "1.3.7.8.9",
        "cps": [
          "https://www.dev.com/repository/"
        ]
      }
    ],
    "authority_info_access": {
      "ocsp_urls": [
        "http://dev.com/ocsp"
      ],
      "issuer_urls": [
        "http://dev.com/ca.crt"
      ]
    },
    "qc_statements": {
      "ids": [
        "0.4.0.1862.1.1",
        "0.4.0.1862.1.3",
        "0.4.0.1862.1.2",
        "0.4.0.1862.1.4",
        "0.4.0.1862.1.5",
        "0.4.0.1862.1.6"
      ],
      "parsed": {
        "etsi_compliance": true,
        "sscd": true,
        "qc_types": {
          "ids": [
            "0.4.0.1862.1.6.1",
            "0.4.0.1862.1.6.2",
            "0.4.0.1862.1.6.3"
          ]
        },
        "limit": {
          "currency": "EURO",
          "amount": 10,
          "exponent": 2
        },
        "pds_locations": [
          {
            "url": "https://dev.com/pds/en/pds.pdf",
            "language": "en"
          }
        ],
        "retention_peiod": 10
      }
    }
  },
  "signature": {
    "signature_algorithm": {
      "name": "SHA256-RSA",
      "oid": "1.2.840.113549.1.1.11"
    },
    "value": "DyWYTIzvGbzIbUbfe0mHSNxjoacuhWWQLhyX/UfHZ5hdYL7osIvwvWu81hGiGK9Eo0279W0YexJ4qp90YIzG8kjKvK7XIbFLmYnCf7+8sXHcvH9w/b8W11cTWmC2+xuYgxCWvHkGwWVpEpZlkDc+YYjilUVQ7wIcingpT98Tek36RIZ8mcNqkXUTrCaWt9Ra4h732J87MSWydifEMYUi+fSKttw7cjBzf6Q5XFkWY0uQtY087EIRldezwa9CZASStaEVzQ7UTsOzLSbmYEQzKK8siO6Z+RjV86ZU47uSZ2cC++/SQBufAQ0Ch/tnN2O3vACfA30QhQblWHe9Wd4taA==",
    "valid": false,
    "self_signed": false
  },
  "fingerprint_md5": "173780910e4430a16d0e7a4ed6667b9f",
  "fingerprint_sha1": "ce69f32ca243df47f6fccb985d6d340b15d5a827",
  "fingerprint_sha256": "a5885ab12e6260345007884baa9f1ca7343719d80381e3a7653be314e80dea34",
  "tbs_noct_fingerprint": "5753c966f66d7db686320109cc4f79553fa25aa45917b8326aab7e9a42507f3b",
  "spki_subject_fingerprint": "af9c1632cdf9bc7b1041ca90006c6956b44cfc0b224ff0c19228b980e8e1b280",
  "tbs_fingerprint": "5753c966f66d7db686320109cc4f79553fa25aa45917b8326aab7e9a42507f3b",
  "validation_level": "unknown",
  "redacted": false
}
```

This also includes the updates to the schema for this PR and #214 and #216


zakird:
This seems fine, though it is a different format than we have certificate policies in. Not sure offhand which is better. Can you have two statements each with ID `0.4.0.1862.1.1` for example?

dadrian:
> This seems fine, though it is a different format than we have certificate policies in.

The thought process here was:
* Indicate the presence of any OID, regardless of if we can parse it or not, to make identifying inclusion easy and consistent (and backwards compatible, as new statements are parsed)
* Provide parsed information that works well for correctly-encoded certificates
* Leave detecting all the "weird" stuff to ZLint, such as detecting the difference between a present array of length zero and a missing array, etc.

> Can you have two statements each with ID 0.4.0.1862.1.1 for example?

I do not believe you can in a valid certificate, but it is possible to encode this.

zakird:
I feel like it might be better to encode in JSON as a list then, similar to how we handle certificate policies? That format doesn't seem to have caused much grief.

dadrian:
The certificate policies all have the same type/body, which is not the case for QCStatements. We would have to either make each field of `qc_statements` parsed an array (a kind of obnoxious option), or create an array of a union type of all possible QCStatements (a really bad option).

Given that duplicate QCStatements would appears in the `ids` field, I'm not sure it's worth making each field in `qc_statements.parsed` an array.

dadrian:
@sleevi What do you think we should do here?

dadrian:
@zakird Certificate policy OIDs correspond to the policy, if I'm not mistaken. Our schema only has two-non-OID fields per policy.

```
CertificatePoliciesData = SubRecordType({
    "id": OID(doc="The OBJECT IDENTIFIER identifying the policy."),
    "cps": ListOf(URL(), doc="List of URIs to the policies"),
    "user_notice": ListOf(UserNoticeData(), doc="List of textual notices to display relying parties."),
})
```

sleevi:
> Can you have two statements each with ID 0.4.0.1862.1.1 for example?

I have yet to find any text that explicitly forbids multiple of the same QC statements. My preferred example is to use the legal liability limit, which allows indicating the currency code of the liability; presumably, one could express multiple currencies (e.g. based on jurisdictional constraints). That said, it also seems very odd/broken to have this, and I’m not sure I haven’t overlooked some prohibition on this.

@dadrian It’s not clear to me how the caller would extract the data if duplicate IDs exist. Would currency code be an array?

dadrian:
> It’s not clear to me how the caller would extract the data if duplicate IDs exist.

That's the exact problem. Right now, you can't. You'd only be able to see the duplicate ID in the list of IDs, but the structures effectively only expose one instance per-ID.

If we want to be able to expose repeated statements, we'd need to make `qc_statements.parsed.limits` an array, as well as `...retention_period`, etc.

The question is: is this worth it?

zakird:
@dadrian how many fields would be in the union? If it's only 5-10, this doesn't really seem like a big issue. It's the same as we do for policies just with 10 fields instead of 2, no?

dadrian:
@zakird Right now, it's about 6. We could adopt the strategy of

```
# Option 1
# Only ID and one of {A, B, C...} are non-null
Statement = SubRecord({
   Id: OID(),
   A: SubRecord()
   B: SubRecord()
   C: SubRecord()
})
QCStatements = ListOf(Statement)
```
I think that the non-null invariant would be worse than just making each parsed statement an array.

```
# Option 2
# In the common case, any non-null field would be an array of length 1
ParsedStatements = SubRecord({
   A: ListOf(SubRecord()),
   B: ListOf(SubRecord()),
   C: ListOf(SubRecord()),
})
QCStatements = SubRecord({
   IDs: ListOf(OID()),
   Parsed: ParsedStatements(),
})
```

Then there's option 3: treat duplicate statement IDs as an error condition and just ensure it is indicated in some way.

My ranking is: 3, 2, 1

zakird:
What's the hesitation with option 1? That's how the extension is structures
in the certificate, consistent with how we parse other fields, and seems
the cleanest.

sleevi:
Considering eIDAS and friends continue to add new `QcStatements`, it’s probably not worth going big-bag-of-structs, because that will only grow over time. Effectively, they use QcStatements as extensions, but without the “at most one” restriction.

> The question is: is this worth it?

Probably not? This seems like a technical edge case, and recall we discussed it some on https://github.com/zmap/zlint/pull/380 because it does seem esoteric, although not forbidden, to repeat statements.

dadrian:
@zakird Option 1 is not consistent with how we do anything.

CertificatePolicies is a single type, defined in RFC 5280, as a sequence of OID, URL, Text, with the URL and text being optional. We expose `parsed.extensions.certificate_policies` as a single array of this type. The OID in a policy does not affect the structure of the data in the policy.

QCStatements are more aligned with extensions, where the OID changes the structure of the data. We expose extensions by uplifting them to scalar fields within a single extensions object. We do not attempt to handle the case where the same extension appears more than once, e.g. a certificate with two subject alternative name extensions. We expose `parsed.extensions.subject_alternative_name` as a single `GeneralName` struct, not as an array of multiple `GeneralName` structs. Similarly, certificate policies is exposed as an array, not an array of arrays.

dadrian:
@zakird I tweaked it to use arrays for each parsed field.

sleevi:
@dadrian Did you also want to include [QcCCLegislation](https://www.etsi.org/deliver/etsi_en/319400_319499/31941205/02.02.03_20/en_31941205v020203a.pdf)

dadrian:
@sleevi Sure, why not! Do you have a cert sitting around that has this field I could use to test?

I came up with the original list by copying https://github.com/zmap/zlint/blob/master/v2/util/qc_stmt.go#L122

sleevi:
> @sleevi Sure, why not! Do you have a cert sitting around that has this field I could use to test?

Sadly, no, I'm not aware of any existing because no other country has legislation in this space like the EU. I only mentioned it because it's part of the spec since this year in case you wanted to be complete.

The other relevant spec is [TS 119 495](https://www.etsi.org/deliver/etsi_ts/119400_119499/119495/01.04.01_60/ts_119495v010401p.pdf), which uses `0.4.0.19495.2` (see Section 5.1) for PSD2 certificates.

PR #222, opened 2020-05-08
tls: Advertise rsa_pkcs1_sha{256,384,512} support by default

codyprime:
When populating for the signature_algorithms extension, `defaultSKXSignatureAlgorithms` is used by default for TLS1.2. We have support for `rsa_pkcs1_sha{256,384,512}`, and some servers
require these to at least be advertised.  Add them in to the default list.

Note: Marked as draft, as the relevant testdata cases need to be updated.


dadrian (approved):
I don't think the transcript tests add _any_ value at this point, and they prevent us from putting in reasonable fixes (such as this) now. I don't anticipate that we have any easy way to regenerate the transcript tests, since they rely on an old version of BoringSSL, so I recommend we delete them, file an issue, and merge this PR. These tests are not reflective of the behavior we want, we should move forward rather than let them continue to prevent merges.

zakird:
I agree with this. @codyprime would you be willing to comment out those tests, and then we can merge?

zakird (approved):

PR #224, opened 2020-06-03
Added explicit check whether two ETSI QCStatements contain an empty statementInfo

mtgag:
This came up while working in https://github.com/zmap/zlint/issues/434. Malformed certificates (nonEmpty statementInfo where this should be empty) can be parsed by zcrypto. 

Adding this to zcrypto would also obsolate a few lints (e_qcstatem_qccompliance_valid, e_qcstatem_qcsscd_valid, ...).

Added certificate that reproduces this behaviour.
[QcStmtEtsiQcComplWithNonEmptyStmtInfoCert19.txt](https://github.com/zmap/zcrypto/files/4724432/QcStmtEtsiQcComplWithNonEmptyStmtInfoCert19.txt)

### DIFF
--- a/x509/qc_statements.go
+++ b/x509/qc_statements.go
@@ -3,6 +3,7 @@ package x509
 import (
 	"encoding/asn1"
 	"encoding/json"
+	"errors"
 )
 
 type QCStatementASN struct {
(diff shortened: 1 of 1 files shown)


zakird:
Thanks, both @mtgag and @sleevi.

PR #230, opened 2020-11-22
CI: Switch to Github Actions.

cpu:
Travis CI has changed their product in such a way that open source projects are seeing long queue times. Since we're already using Github it makes sense to use Github actions for CI. This removes an external dependency and seems like a more modern approach overall.

Two new Github actions workflows are added to replace the Travis config:

1. `go.yml` - this builds and runs unit tests under Go 1.14. We can add 1.15 later.
1. `gofmt.yml` - this checks that all files produce no diffs under `gofmt -l -s`.

Notably I haven't ported over the `slack` notification from the Travis config. I have no idea what Slack org/channel this webhook config was for. I would need someone familiar with the old config to find a way to replace it with Github actions.

Once this is merged a maintainer will have to change the `master` branch protection config to remove Travis and add the two new actions.


cpu:
> Notably I haven't ported over the slack notification from the Travis config. I have no idea what Slack org/channel this webhook config was for. I would need someone familiar with the old config to find a way to replace it with Github actions.

@zakird @dadrian Do either of you know about this? Is it important to port forward for this repo?

dadrian:
It is not important to port the Slack notification.

.github/workflows/go.yml
```diff
@@ -0,0 +1,25 @@
+name: Go
+on: [push, pull_request]
+jobs:
+
+  build:
+    name: Build and Unit Test
+    runs-on: ubuntu-latest
```
dadrian, line 7: I'm not as familiar with Github Actions as other CI, but we could probably use a less-heavy runner, something more akin to the `golang` containers on Dockerhub.

dadrian (approved):

.github/workflows/go.yml
cpu: > I'm not as familiar with Github Actions as other CI, but we could probably use a less-heavy runner, something more akin to the golang containers on Dockerhub.

I'm also super-duper new to Github Actions but it seems like there aren't any light-weight options for the `runs-on` parameter based on [the docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on). FWIW in practice I'm seeing these actions run must faster than the Travis equivalents :shrug:

sleevi: @cpu If there’s a Dockerhub image that does the needful, it’s also real easy to use a Docker step, which will mount the workspace into the image.

My terribly-hacky still WIP [cabforum integration](https://github.com/sleevi/cabf-scratch/blob/main/.github/workflows/build-docs.yml) does this for the Pandoc step, and hopefully next week will have a custom step that builds all the tools as a Docker image to begin with.

cpu: I'm largely following the [setup-go](https://github.com/actions/setup-go) instructions. I don't believe there's a more tailored docker image to use for this but would be open to pointers if folks know of one.

cpu:
I think the Travis config is busted - seems blocked on waiting for a status to be reported. If this looks OK to folks I'd appreciate if an admin could merge skipping that status.

cpu:
@dadrian Thanks for the merge. Can you also update the `master` branch protections to remove Travis and enforce the new actions? I don't have the permissions for this repo. I just did a quick test and it looks like the old settings are still in place:
Screen Shot 2020-11-23 at 3 03 53 PM

zakird:
I think I've set this. @cpu can you check?

cpu:
@zakird Thanks! The new settings look correct but for whatever reason my [test PR with bad gofmt'ing](https://github.com/zmap/zcrypto/pull/231/files) didn't get blocked. I'll debug that action separately.

Issue #232, opened 2020-11-23
Fix gofmt github action


cpu:
A [test PR](https://github.com/zmap/zcrypto/pull/231) with bad `gofmt`ing wasn't blocked by the Github action added in https://github.com/zmap/zcrypto/pull/230/

PR #233, opened 2020-11-23
CI: Fix `gofmt` Github Action.

cpu:
This updates the gofmt action added in https://github.com/zmap/zcrypto/pull/230 and fixes https://github.com/zmap/zcrypto/issues/232. A few changes:

1. Use the most recent version of the action published, add a note about updating to a new tag to support using `-s` with `gofmt` in the future. (See https://github.com/Jerome1337/gofmt-action/issues/5). In the meantime we can enforce `gofmt` without the `-s` mode.

2. Actually check out the code before trying to `gofmt` it. Otherwise everything will pass... This was my bad. I'm still getting adjusted to the Github actions way of doing things.

Verified this works correctly with [a branch](https://github.com/zmap/zcrypto/compare/master...cpu:cpu-gofmt-testing) on my fork. Here's the [failed Gofmt action run](https://github.com/cpu/zcrypto/runs/1444609999?check_suite_focus=true).

### DIFF
--- a/.github/workflows/gofmt.yml
+++ b/.github/workflows/gofmt.yml
@@ -5,10 +5,17 @@ jobs:
   build:
     name: Gofmt check
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go: [ '1.14' ]
     steps:
 
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    # TODO(@cpu): Switch to v1.0.4 and add `gofmt-args: -l -s` once new release
+    # is available. See https://github.com/Jerome1337/gofmt-action/issues/5
     - name: Check gofmt
-      uses: Jerome1337/gofmt-action@v1.0.2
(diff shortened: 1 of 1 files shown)


cpu:
@zakird Sorry for the trouble but there might be something up with the branch protection settings. The gofmt action completed but it isn't marked as required, instead some other ghost check is marked required and stuck waiting for status:
settings

cpu:
> The gofmt action completed but it isn't marked as required, instead some other ghost check is marked required and stuck waiting for status:

Ahh, I think I caused this when I added an explicit Go version matrix to the check to make sure the formatting was consistent. The protection check changed names I think. Pretty annoying that it's so weakly attached to the repo config.... Would you mind twiddling?

zakird:
This should be fixed. Also bumped your access on the repository.

zakird (approved):

cpu:
> This should be fixed. Also bumped your access on the repository.

Thanks!

PR #235, opened 2020-11-28
chore: update to Go 1.15, bump some deps.

cpu:
Switches CI and go.mod to Go 1.15. Also updates three dependencies that
were easy to upgrade in place:

* github.com/weppos/publicsuffix-go
* golang.org/x/crypto
* golang.org/x/net


cpu:
@zakird Do you mind stamping this too if it looks OK to you? Also I think I still can't tweak branch protection settings with my access level. If you could tweak them for the 1.15 versions that would be great. I promise this is the last time for a while :laughing: Thanks for your help.

zakird (approved):

cpu:
Thank you!

Issue #236, opened 2020-12-06
Default cipher-suite list for zcrypto and exposing PortableCiphers list


mzpqnxow:
Hello,

Recently, I looked at an issue a user entered in zcrypto (#229) as well as zgrab2 [281, here](https://github.com/zmap/zgrab2/issues/281) which was a failure of `zgrab2 http --use-ssl` to negotiate a cipher-suite with a specific device the user included for testing. The failure occurred when using the zgrab2/zcrypto "default" cipher-suite

I determined the "fix" was very easy and that there was not necessarily anything wrong with the device or zgrab2/zcrypto as I initially suspected. The device supported TLS1.2 and only *two* specific cipher-suites- nothing else. Neither cipher-suite was included in the default list in zcrypto- though both *are* available/implemented in zcrypto. Turns out the default list is rather conservative compared with what I thought

The two ciphers were:

```
	TLS_RSA_WITH_AES_256_CBC_SHA256               = 0x003D
...
	TLS_RSA_WITH_AES_256_GCM_SHA384               = 0x009D
```

There are a few ways to work around this issue if you can consider it an issue, I guess it's arguably just pilot error-

```
$ echo 1.179.232.207 | zgrab2 http -p 1080 --use-https --max-redirects=5 --cipher-suite=3d
$ echo 1.179.232.207 | zgrab2 http -p 1080 --use-https --max-redirects=5 --cipher-suite=9d
```

Or, because the `SafariCiphers` list happens to have `TLS_RSA_WITH_AES_256_CBC_SHA256`, the following would work as well:

```
echo 1.179.232.207 | zgrab2 http -p 1080 --use-https --max-redirects=5  --cipher-suite=safari-only
```

I initially suspected the device was using a very obscure cipher because it's a NETGEAR router- but when I saw that it was a pretty modern cipher, and TLS1.2, I decided I wanted to look into it a little as I use zgrab2 regularly to gather SSL/TLS metadata and was wondering how many handshake errors that I was having (and not investigating) were due to the cipher-suite list being too limited- I had made assumptions about the default cipher-suite list that turned out to be incorrect. Oops!

When I looked at how the `--cipher-suite` flag worked, I saw that it was a map to various cipher-suite lists hard-coded into zcrypto, e.g. [SafariCiphers](https://github.com/zmap/zcrypto/blob/3719af1573cf8a5e8cfad555065be8b00a5a6d8e/tls/cipher_suites.go#L1031) for safari-only

[...]

Issue #229, opened 2020-11-16
why handshake failure ?


chushuai:
Can be accessed normally through the browser

https://1.179.232.207:1080

Here is the test code
```
import (
	"fmt"
	"github.com/zmap/zcrypto/tls"
	"testing"
)

func TestTLS(t *testing.T) {
	config := tls.Config{
		InsecureSkipVerify: true,
		MaxVersion:         tls.VersionTLS12,
	}
	/*
	https://1.179.232.207:1080
	*/
	conn, err := tls.Dial("tcp", "1.179.232.207:1080", &config)
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(conn)
	defer conn.Close()
}

```

mzpqnxow:
This seems to be the same as https://github.com/zmap/zgrab2/issues/281, you can read what I wrote there

I wonder if the default cipher-suite profiles need to be updated in zcrypto to be more inclusive to junk like this device (which seems to be an old NETGEAR router)

chushuai:
I don't have this problem with the latest version of golang crypto/tls.  zcrypto's crypto/tls is too old version
@cpu @mzpqnxow

mzpqnxow:
@chushuai are you building from source, or using a release package from somewhere?

mzpqnxow:
@chushuai if you would like to test, I have a zcrypto branch @ https://github.com/mzpqnxow/zcrypto/tree/feature/portable-cipher-suite-profile that provides a pre-build list of cipher-suites that will address this and should handle any similar issues you run into with other endpoints that are picky. If you use that branch, you'd need to update your code to specify `CipherSuites` in the TLS config:

	`CipherSuites = tls.PortableCiphers`

I sent in a PR for this, but please feel free to test in the mean time. I also tested it with `zgrab2` and will add a branch that makes use of this once/if this is merged

mzpqnxow:
Was this issue resolved for you @chushuai ? A lot of work has been done in zcrypto since the time you created this

mzpqnxow:
> Was this issue resolved for you @chushuai ? A lot of work has been done in zcrypto since the time you created this

Pretty sure this is resolved and can be closed

PR #238, opened 2020-12-07
Add a "PortableCipher" list that aims to be as compatible as possible to supplement the browser cipher lists

mzpqnxow:
See #229 and for more detail, see #236

This is a very small and non-invasive patch that adds an additional static list of ciphers that are already implemented. The objective is to provide an easily referenced collection of cipher-suites for applications using zcrypto that is as compatible as possible given the current set of algorithms implemented in zcrypto.

This should have no impact on current users of zcrypto. If this is merged, it can be used by zgrab2 to more easily access a list of ciphers focused on portability via, e.g. `--cipher-suite=portable`, which would be added to the other existing options in the zgrab2 http module (`firefox-only`, `chrome-only`, `safari-only`, etc.)

The original "default" is kept the same. Also, in this new `PortableCiphers` list, for the majority of cases, the same cipher will be negotiated as would be when using the default as I started the list with the same elements in the default list, and then extended it by adding the additional ciphers, to minimize the risk of any unexpected issues for those that choose to use this over the default.

Against a specific real-world dataset that I work with regularly, which consists of ~50k HTTPS endpoints, there's a 2% increase in the amount of successful SSL/TLS negotiations compared with the default cipher list. This is relatively modest, but important depending on the use-case.

I also added a single cipher (`TLS_RSA_WITH_AES_256_GCM_SHA384`) to the `ChromeCiphers` list in this commit- this may not be appropriate, if it needs to be removed I can remove it- or feel free to do it in-place in the GH PR interface. The `ChromeCiphers` list is currently out of date anyway, so maybe there's not much harm leaving it in?

Thanks for your consideration!


tls/cipher_suites.go
```diff
@@ -1101,6 +1102,75 @@ var SafariNoDHECiphers []uint16 = []uint16{
...
+	TLS_DH_ANON_EXPORT_WITH_RC4_40_MD5,
+	TLS_DHE_DSS_EXPORT_WITH_DES40_CBC_SHA,
+	TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA,
+	TLS_DHE_DSS_WITH_AES_128_CBC_SHA256,
+	TLS_DHE_DSS_WITH_AES_128_GCM_SHA256,
+	TLS_DHE_DSS_WITH_AES_256_CBC_SHA,
+	TLS_DHE_DSS_WITH_AES_256_CBC_SHA256,
+	TLS_DHE_DSS_WITH_AES_256_GCM_SHA384,
+	TLS_DHE_DSS_WITH_DES_CBC_SHA,
+	TLS_DHE_DSS_WITH_RC4_128_SHA,
```
dadrian, line 1163: I don't think the `ANON` or `DSS` suites are actually capable of completing a handshake.

tls/cipher_suites.go
```diff
@@ -1101,6 +1102,75 @@ var SafariNoDHECiphers []uint16 = []uint16{
...
+	TLS_ECDH_ECDSA_WITH_AES_128_CBC_SHA,
+	TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
+	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+	TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+	TLS_ECDH_ECDSA_WITH_3DES_EDE_CBC_SHA,
+	TLS_ECDH_RSA_WITH_AES_256_CBC_SHA384,
+	TLS_ECDH_RSA_WITH_AES_128_CBC_SHA256,
+	TLS_ECDH_RSA_WITH_AES_256_CBC_SHA,
+	TLS_ECDH_RSA_WITH_AES_128_CBC_SHA,
+	TLS_ECDH_RSA_WITH_3DES_EDE_CBC_SHA,
```
dadrian, line 1142: I don't think we actually support the non-ephemeral DH and ECDH suites.

tls/cipher_suites.go
```diff
@@ -1101,6 +1102,75 @@ var SafariNoDHECiphers []uint16 = []uint16{
...
+	TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,
+	TLS_RSA_WITH_AES_256_GCM_SHA384,
+	TLS_RSA_WITH_AES_128_GCM_SHA256,
+	TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384,
+	TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,
+	TLS_DHE_DSS_WITH_AES_128_CBC_SHA,
+	TLS_ECDH_ECDSA_WITH_AES_256_CBC_SHA384,
+	TLS_ECDH_ECDSA_WITH_AES_128_CBC_SHA256,
+	TLS_ECDH_ECDSA_WITH_AES_256_CBC_SHA,
+	TLS_ECDH_ECDSA_WITH_AES_128_CBC_SHA,
```
dadrian, line 1133: I don't think we actually support the non-ephemeral DH and ECDH suites.

tls/cipher_suites.go
```diff
@@ -1101,6 +1102,75 @@ var SafariNoDHECiphers []uint16 = []uint16{
...
+	// Most of the other implemented ciphers, in a somewhat reasonable order
+	TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384,
+	TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,
+	TLS_ECDHE_ECDSA_WITH_3DES_EDE_CBC_SHA,
+	TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,
+	TLS_RSA_WITH_AES_256_GCM_SHA384,
+	TLS_RSA_WITH_AES_128_GCM_SHA256,
+	TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384,
+	TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,
+	TLS_DHE_DSS_WITH_AES_128_CBC_SHA,
```
dadrian, line 1129: I'm not sure that we actually can complete a handshake that uses `DSS` certs.

dadrian (approved):
LGTM % notes on a few specific ciphers.

zakird:
This looks good to me once @dadrian's concerns are addressed. I agree that we shouldn't advertise ciphers we're unable to handshake for.

tls/cipher_suites.go
mzpqnxow: Hmm good point, I didn't actually test with endpoints preferring each. I just took from the list of implemented ciphers in that file. I'll rig something quick up and confirm that what you're saying is correct (and probably more important, confirm I didn't specify any others that won't actually work if they get chosen)

tls/cipher_suites.go
mzpqnxow: Yeah, I've never actually even tried to use an ANON cipher-suite. Same as the comment below, I'll do testing across all of these and see what shakes out before reporting back with some additional commits with VERIFIED completed handshakes for each. Sorry, this was really careless.

mzpqnxow:
Yeah, apologies for not actually testing to ensure that the handshakes can complete. I jumped to conclusions based on the implemented ciphers list. I'll confirm all of them before coming back :)

dadrian:
@mzpqnxow It's not unreasonable to expect the implemented ciphers list to contain ciphers that are implemented! 😆

zakird:
Hey @mzpqnxow, I'd absolutely love to get this merged. Any chance you'd be willing to take a look at that list of ciphers?

mzpqnxow:
Hey @zakird yes, I’m sorry this fell off of my map (or maybe I fell off of the map) I’d like to get to this in the upcoming week if you’re still interested

zakird:
Absolutely!

tls/cipher_suites.go
mzpqnxow: @dadrian you are correct about DSS certs- so I will remove those (for starters) I still need to test the non-ephemeral cipher-suites you mentioned- and the ANON ones..  in the process of doing that now.

```
openssl.old s_server -dcert server2.pem -dkey server2.pem -cert server1.pem -key server1.pem  -cipher DH-DSS-AES256-GCM-SHA384:DHE-DSS-AES256-GCM-SHA384:DHE-DSS-AES256-SHA256:DH-DSS-AES256-SHA256:DHE-DSS-AES256-SHA:DH-DSS-AES256-SHA:DHE-DSS-CAMELLIA256-SHA256:DH-DSS-CAMELLIA256-SHA256:DHE-DSS-CAMELLIA256-SHA:DH-DSS-CAMELLIA256-SHA:SRP-DSS-AES-128-CBC-SHA:DH-DSS-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:DHE-DSS-AES128-SHA256:DH-DSS-AES128-SHA256:DHE-DSS-AES128-SHA:DH-DSS-AES128-SHA:DHE-DSS-CAMELLIA128-SHA256:DH-DSS-CAMELLIA128-SHA256:DHE-DSS-SEED-SHA:DH-DSS-SEED-SHA:DHE-DSS-CAMELLIA128-SHA:DH-DSS-CAMELLIA128-SHA:DHE-DSS-RC4-SHA:SRP-DSS-3DES-EDE-CBC-SHA:EDH-DSS-DES-CBC3-SHA:DH-DSS-DES-CBC3-SHA -tls1_2 -debug -msg -www
```
...

```
{"ip":"127.0.0.1","data":{"http":{"status":"unknown-error","protocol":"http","result":{},"timestamp":"2021-02-16T19:23:24-05:00","error":"tls: server's certificate contains an unsupported type of public key: *dsa.PublicKey"}}}
```

tls/cipher_suites.go
mzpqnxow: Confirmed, no non-ephemeral DH are supported. Working on the commit with all of the unsupported stuff removed

tls/cipher_suites.go
mzpqnxow: Confirmed, can't complete a handshake with DSS certs, upcoming commit will reflect that

mzpqnxow:
@zakird @dadrian I just pushed a rather large commit onto the existing PR...

Unfortunately, I did this all as one commit, making it pretty hard for you to review as individual sections. If you'd prefer, I can revert to the old version and include *only* the removal of the unsupported suites from the `PortableCiphers` list that I added.

What is in this:

1. Fixup to the `PortableCiphers`- commenting out the unsupported suites
2. Fixups for any of the browser-specific suite lists, commenting out unsupported suites- in one or two cases, there were DSS or non-ephemeral DH ciphers that were listed above a few others. I don't know how you feel about that- if the goal is to look like those browsers over the wire and you're OK with the trade-off of potentially losing a few handshakes then you probably won't want this. Well, you get the idea, let me know if you prefer I uncomment them and leave them as is
3. Annotations (comments) but not removal in a few of the suite lists that include unsupported suites. This is also easy for me to undo if you don't want the clutter all over the place, just let me know

If you get a chance, just let me know at a high level which (if any) of the three categories of changes made should be removed. Sorry I went a bit overboard but I figured since I was putting the effort in I might as well be as verbose and correct as possible. Not everyone agrees on that though :)

BTW- this came out of this effort... maybe useful in the future (but probably not): https://gist.github.com/mzpqnxow/b863ee8ea4b30c96c10f7ee4d101e887

PR #239, opened 2020-12-11
URIs option missing from Name Constraints 

ZsofiaTomicsko:
Hi!

I couldn't find an option to use URIs for name constraints inside x509.go, when it should be possible according to https://tools.ietf.org/html/rfc5280#section-4.2.1.10 .


zakird (approved):

zakird:
Looks great! Thank you!

PR #240, opened 2020-12-21
CI: automate PRs for weppos/publicsuffix-go updates.

cpu:
The upstream `github.com/weppos/publicsuffix-go` data updates with frequency. Since release tags are not published as often as the data updates are merged and the API is very stable it makes sense for ZCrypto to track upstream master.

Manually updating the dependency when `publicsuffix-go` is updated is a drag. This commit adds a workflow that can open the update PRs automatically similar to the way ZLint updates gTLD data.

### DIFF
--- a/.github/workflows/psl-update.yml
+++ b/.github/workflows/psl-update.yml
@@ -1,0 +1,53 @@
+name: psl-update
+on:
+  schedule:
+    # Run every hour, at the 15 minute mark. E.g.
+    # 2020-11-29 00:15:00 UTC, 2020-11-29 01:15:00 UTC, 2020-11-29 02:15:00 UTC
+    - cron:  '15 * * * *'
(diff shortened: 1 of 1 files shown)


cpu:
Once this is merged a bot PR will bump the dep from v0.13.0 (from Feb) to the latest commit (~27d old). Pulling that update into ZLint will resolve a complaint about inaccurate lint results from stale PSL data: https://github.com/zmap/zlint/issues/523

zakird (approved):

zakird:
I'm a little bit scared that we could break something by tracking their master, but things are currently in a broken state and there isn't a ton we can do if they're not doing releases, so 🤞 .

cpu:
@zakird :+1: definitely open to revisiting if it ends up being problematic.

PR #249, opened 2021-01-25
Created CRLSet and OneCRL packages

dissoupov:


cpu:
@dissoupov Is there some additional context for this change? Why is it needed, etc? How does it differ from the code in https://github.com/zmap/zcrypto/tree/master/x509/revocation that already derives from agl's work for CRLSets? The new `github.com/mozilla/OneCRL-Tools` dependency also has a pretty big transitive footprint :-/

dissoupov:
@cpu looks like I've done some redundant work as per @zakird   request.
let me check, if we can simply use the existing code.

cpu:
> let me check, if we can simply use the existing code.

@dissoupov Ok! I think some of it is a little crusty. It's only something I know about from chasing down a test failures in https://github.com/zmap/zcrypto/commit/cd01ac2110c6b10dbbd94da2382cb19e4e25e055

If you've already chatted with @zakird about this I'll bow out to let you folks manage this one. I thought it was a PR out of the blue from the community. Feel free to tag me if you need another reviewer later on.

zakird:
@cpu Yeah, this one is coming from me. This code is supposed to add support for pulling CRL data from OneCRL and CRLSets. If we already have code, then we should certainly clean up and not have duplicate. I didn't know we had anything in the repo, good catch!!  There's also a question of whether we should move this into a different repo, or keep it here. I don't have a strong opinion either way.

cpu:
> There's also a question of whether we should move this into a different repo, or keep it here. I don't have a strong opinion either way.

The existing Mozilla OneCRL code in the repo seems pretty lean. My only hesitance with putting more revocation checking code into the ZCrypto repo is if it requires a lot of 3rd party code.

dissoupov:
@cpu @zakird after analyzing the existing code I think it makes sense to close this PR and improve the existing code:
- make `Check` function consistent in both packages by providing the issuer cert as parameter:
`Check(cert, issuer *x509.Certificate) *Entry`
- add `FetchAndParse()` to `mozilla` also as it only provides `Parse` method.
- introduce a factory pattern to provide a custom URL to fetch from (from this PR),
 and keep existing `FetchAndParse` to use a default factory.

PR #258, opened 2021-02-03
Add CRLSet and ONECrl options to Verifier

dissoupov:


verifier/verifier_test.go
```diff
@@ -277,10 +306,104 @@ func (vt *verifyTest) checkVerifyResult(res *VerificationResult) error {
...
+		Intermediates: []string{
+			data.PEMLEX3SignedByDSTRootCAX3, // idx=1
+			data.PEMLEX3SignedByISRGRootX1,
+		},
+		Roots: []string{
+			data.PEMDSTRootCAX3SignedBySelf, // idx=3
+		},
+		CurrentTime: 1501804800, // 2017-08-04T00:00:00
+		ExpectedChains: [][]int{
+			{0, 1, 3},
```
dadrian, line 330: It looks like we still add chains that are rejected by a CRLSet / OneCRL to `ExpectedChains`. Is that intended behavior?

dadrian (commented):
This looks really good overall, I just have one question about behavior. I especially appreciate the tests.

verifier/verifier_test.go
dissoupov: @dadrian that is consistent with existing behaviour when a cert is revoked via CRL or OCSP.
I checked `OCSPRevoked` and `CRLRevoked` , and they do not affect currently `ExpectedChains` if a cert is revoked.

I guess, if we decide to change this behaviour, it should be done in a separate PR along with `OCSPRevoked` and `CRLRevoked`.

@zakird FYI
let's discuss

x509/revocation/mozilla/mozilla.go
```diff
@@ -72,8 +74,9 @@ type Entry struct {
 
 // SubjectAndPublicKey specifies a revocation entry by Subject and PubKeyHash
 type SubjectAndPublicKey struct {
```
bwireman, line 76: Is this going to cause issues for users, since we're heavily changing an exported strucy?

verifier/verifier.go
```diff
@@ -212,6 +217,18 @@ func (v *Verifier) Verify(c *x509.Certificate, opts VerificationOptions) (res *V
 		res.Parents = parentsFromChains(res.CurrentChains)
 	}
 
+	if opts.OneCRL != nil && opts.OneCRL.Check(c) != nil {
+		res.InRevocationSet = true
+	}
+	if opts.CRLSet != nil {
```
bwireman, line 223: I don't know how expensive these checks will end up being, but I'm guessing not super cheap, so maybe we should use `else if` since theres nothing to be gained from doing both branches

```suggestion
	else if opts.CRLSet != nil {
```

x509/revocation/mozilla/mozilla.go
dissoupov: yes, if they change the commit.
but the existing users are already broken or not using the latest OneCRL as the schema has changed and the current implementation fails to parse the most recent CRL set.

verifier/verifier.go
dissoupov: `else` is not appropriate here.
`CRLSet` and `OneCRL` are different source of info. and it's already in memory. no penalty on the check.

bwireman: Sure but we don't get anything out of doing the second check if the first branch was true because they set the same bool?

dissoupov: @bwireman instead of `else` I added this condition `!res.InRevocationSet` 

`if !res.InRevocationSet && opts.CRLSet != nil {`

verifier/verifier_test.go
dissoupov: In the code, the chains are built before the revocation checks are performed:
```
	allChains = append(allChains, res.CurrentChains...)
	allChains = append(allChains, res.ExpiredChains...)
	allChains = append(allChains, res.NeverValidChains...)

        if opts.OneCRL != nil {}
        if opts.ShouldCheckOCSP {}
        if opts.ShouldCheckCRL {}
```

zakird (approved):

bwireman (approved):

dissoupov:
@zakird @dadrian please merge

PR #270, opened 2021-03-15
Vendoring crypto/dsa from go 1.16

### LINKED ISSUE #269: Vendor crypto/dsa from Go 1.16
zlint is facing issues with a third party linter that does not like that we are referring to this package, so let's a vendor a copy.

christopher-henderson:
I believe the interesting thing of note is that `crypto/dsa` makes reference to `crypto/internal/randutil`, so I went ahead and mimicked this `internal` package structure.

Resolves #269


dsa/dsa.go
```diff
@@ -0,0 +1,309 @@
+// Copyright 2011 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package dsa implements the Digital Signature Algorithm, as defined in FIPS 186-3.
+//
+// The DSA operations in this package are not implemented using constant-time algorithms.
+//
+// Deprecated: DSA is a legacy algorithm, and modern alternatives such as
```
cpu, line 9: I think we'll want to drop the `Deprecated:` bit here to avoid further `staticcheck` results. Maybe use `Warning:` instead?

cpu (changes requested):
Looks OK to me modulo not deprecating the vendored version.

dsa/dsa.go
christopher-henderson: Huh, I admit that I didn't know that [godoc](https://github.com/golang/go/wiki/Deprecated) picks up on that (I've honestly paid more attention to Rust documentation capabilities over the past several years.).

cpu: It was a TIL for me too. It seems underdocumented! The best source I could find was one small sentence from https://blog.golang.org/godoc

> To signal that an identifier should not be used, add a paragraph to its doc comment that begins with "Deprecated:" followed by some information about the deprecation

cpu: >  I didn't know that [godoc](https://github.com/golang/go/wiki/Deprecated) picks up on that 

That's a better reference, ty!

cpu (approved):

PR #274, opened 2021-03-27
Use the internal DSA package throughout

christopher-henderson:
The previous change to vendor the `crypto/dsa` package did not update all code points that use that package, which meant that, for example. `zcrypto/x509` would leak `crypto/dsa`. This would, in turn, cause `*dsa.PublicKey is not *dsa.PublicKey` type errors because they are the same name from different scopes.


cpu (approved):
Looks good to me. Thanks @christopher-henderson

PR #277, opened 2021-04-13
Check revocation on non expired cert only

dissoupov:


verifier/verifier.go
```diff
@@ -249,37 +249,41 @@ func (v *Verifier) VerifyWithContext(ctx context.Context, c *x509.Certificate, o
...
-	if !res.InRevocationSet && opts.CRLSet != nil {
-		for _, parent := range res.Parents {
-			if opts.CRLSet.Check(c, hex.EncodeToString(parent.SPKIFingerprint)) != nil {
-				res.InRevocationSet = true
-				break
+		//
+		// Check revocation on non expired cert only
+		//
+
+		if opts.OneCRL != nil && opts.OneCRL.Check(c) != nil {
```
bwireman, line 257: Should it be a flag if we check revocation of expired flags or not?

dissoupov: no, if the cert is expired, it's already invalid and can be removed from CRL and not served by OCSP responders.

dissoupov:
@bwireman A certificate that expires does not go onto the CRL because an expired certificate is automatically rejected...

Some description about revocation: https://www.securew2.com/blog/certificate-revocation-crl-explained

sleevi:
> @bwireman A certificate that expires does not go onto the CRL because an expired certificate is automatically rejected...

This is not quite correct. Both OCSP and CRLs can continue to provide information about expired certificates. This is dependent on the CA and their policies, as well as the intended use case of the certificate.

For example, a certificate is issued at T=0 and expires at T=10, and is used to sign an document at T=5, as well as timestamped by a timestamping authority (establishing the time). At T=11, it is discovered that the private key was compromised at T=4. The CA MAY decide to provide revocation information via CRL and/or OCSP, in order to provide information about whether the signature created at T=5 was/is valid.

This is intrinsic to the design of both CRLs and OCSP. Yes, if you attempt to verify the certificate with the validity time of T=11, the certificate will fail (expired), but if at T=11 you attempt to verify it with the input validity time set at T=5 (the countersigned timestamp time), you may still be fetching a **fresh** OCSP response (at T=11) for the expired certificate.

This is an example of why both the CRL and OCSP specifications explicitly don’t assume expired certificates are removed. However, this varies on the PKI and use case, as other PKIs (e.g. authentication PKIs, TLS PKIs) don’t have the issue of “asking about the past”, because they are real-time protocols, and so there, it’s typically assumed that expired certificates will be “eventually” removed. Even then, however, there may be restrictions on when it is removed (e.g. one full publication cycle of the repository artifacts, such as if your CRL validity is 7 days, it may be 14 days before being removed from CRLs).

I’m not sure to what extent this matters for this specific case, but in general for PKI, one cannot simply assume expired means “skip revocation checks”; it’s dependent on use case.

dissoupov:
@sleevi in reality we see a lot of errors returned from OCSP on expired certs.
Your example for Code Signing certs is a valid case, but I was under impression that ZMap is intended to scan TLS certs that are anyhow invalid beyond expiration.

Do you have a proposal for an alternative behavior?  Introducing yet another flag?

dissoupov:
@sleevi  @bwireman we're closing this for now, and will disable revocation flags on expired certs before calling `Validate`

PR #279, opened 2021-04-16
Porting ocsp package from the latest standard lib

dissoupov:
A lot of OCSP responders return `400 Bad request` or `unauthorized`.
Building with the latest `ocsp` package from the standard lib, seems to be fixing the problem.

### DIFF
--- a/x509/revocation/ocsp/ocsp.go
+++ b/x509/revocation/ocsp/ocsp.go
@@ -1,11 +1,26 @@
+// Copyright 2013 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package ocsp parses OCSP responses as specified in RFC 2560. OCSP responses
+// are signed messages attesting to the validity of a certificate for a small
+// period of time. This is used to manage revocation for X.509 certificates.
 package ocsp
 
 import (
 	"crypto"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	_ "crypto/sha1"
+	_ "crypto/sha256"
+	_ "crypto/sha512"
 	"encoding/asn1"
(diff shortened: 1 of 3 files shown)


dissoupov:
@dadrian @casebenton @zakird could you please take a look if it makes sense to backport standard `ocsp` package?
I see you did add this one 3 years ago and I'm not sure if there were some modifications in functionality, or it was simply copied.

Reason for the update:
- A lot of OCSP responders return 400 Bad request or unauthorized.
Building with the latest ocsp package from the standard lib, seems to be fixing the problem.
- The latest package allows to specify options for the request

@codyprime @bwireman FYI

x509/revocation/ocsp/ocsp.go
```diff
@@ -1,173 +1,364 @@
...
+	case TryLater:
+		return "try later"
+	case SignatureRequired:
+		return "signature required"
+	case Unauthorized:
+		return "unauthorized"
+	default:
+		return "unknown OCSP status: " + strconv.Itoa(int(r))
 	}
-	return json.Marshal(&aux)
```
dadrian, line 51: We vendored and edited it to add JSON support.

dadrian (commented):
I'm fine updating the package, but the reason we backported it was to add JSON support. I don't know that anyone was using it, so if we have another way of doing JSON for OCSP, I'm fine (even prefer!) dropping the vendor.

x509/revocation/ocsp/ocsp.go
dissoupov: @dadrian I can try to restore JSON support if it's really needed.
what is the scenario for JSON in OCSP?

dadrian: I believe it was done for data collection purposes for bulk revocation. I don't know that revocation ever came to fruition outside of some Censys prototypes, so it's probably fine to drop.

dadrian (approved):

PR #284, opened 2021-05-03
backporting asn1, pkix to allow permissive parsing

dissoupov:
x509.ParseCertificate fails on many certs in `censys` pipeline.
The parsing should be permissive and allowing, for example, BER encoded certs or not strictly DER encoded.

This change back ports the latest `pkix` and `asn1` packages and disables some strict validation in decoding and parsing certificates.


x509/x509.go
```diff
@@ -1907,25 +1942,27 @@ func parseCertificate(in *certificate) (*Certificate, error) {
...
+								}
 							}
 						}
 						if qualifier.PolicyQualifierId.Equal(cpsURIOID) {
 							var cpsURIRaw asn1.RawValue
-							if _, err = asn1.Unmarshal(qualifier.Qualifier.FullBytes, &cpsURIRaw); err != nil {
-								return nil, err
+							if _, err = asn1.Unmarshal(qualifier.Qualifier.FullBytes, &cpsURIRaw); err == nil {
+								out.CPSuri[i] = append(out.CPSuri[i], string(cpsURIRaw.Bytes))
 							}
```
aspacewalz, line 1965: Do we want this to also include a check for `AllowPermissiveParsing`?

aspacewalz (commented):
This looks good to me. I appreciate the tests for the permissive parsing.

dissoupov:
yes, I'm fixing that

x509/x509.go
dissoupov: fixed

aspacewalz (approved):

PR #285, opened 2021-05-10
forking from golang.org/x/crypto/cryptobyte to allow permissive parsing

dissoupov:


bwireman (approved):
Looks the same to me (ignoring import diffs) as 
https://github.com/golang/crypto/tree/internal-branch.go1.16-vendor/cryptobyte @ 06a226fb4e3765ef3f48aa2852b401bc7b98e981

PR #292, opened 2021-05-27
Fix Name.String() to legacy behavior, permissive parsing asn1.IA5String

dissoupov:


dissoupov:
Censys needs a legacy DN representation.
The standard X509 prints Name in reverse order:

```
expected: "SERIALNUMBER=67890, CN=name, SERIALNUMBER=12345, C=US, C=RU, POSTALCODE=48109, ST=Michigan, L=Ann Arbor, STREET=2260 Hayward St, O=University of Michigan, OU=0x21, CN=common",
legacy:   "CN=common, OU=0x21, O=University of Michigan, street=2260 Hayward St, L=Ann Arbor, ST=Michigan, postalCode=48109, C=US, C=RU, serialNumber=12345, CN=name, serialNumber=67890"
```

This change introduces a conditional flag that will allow legacy behavior

x509/pkix/pkix.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/zmap/zcrypto/encoding/asn1"
 )
 
+// LecacyNameString allows to specify legacy ZCrypto behaviour
+// for X509Name.String() in reverse order
+var LecacyNameString = false
```
codyprime, line 20: s/Lecacy/Legacy

encoding/asn1/asn1.go
```diff
@@ -468,10 +468,12 @@ func isPrintable(b byte, asterisk asteriskFlag, ampersand ampersandFlag) bool {
 // parseIA5String parses an ASN.1 IA5String (ASCII string) from the given
 // byte slice and returns it.
 func parseIA5String(bytes []byte) (ret string, err error) {
-	for _, b := range bytes {
-		if b >= utf8.RuneSelf {
-			err = SyntaxError{"IA5String contains invalid character"}
-			return
+	if !AllowPermissiveParsing {
```
codyprime, line 471: Am I correct in assuming these changes in `asn1.go` are unrelated to the PR description?

dissoupov: it was a pending previous PR

dissoupov: updated commit info

x509/pkix/pkix.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/zmap/zcrypto/encoding/asn1"
 )
 
+// LegacyNameString allows to specify legacy ZCrypto behaviour
+// for X509Name.String() in reverse order
+var LegacyNameString = false
```
codyprime, line 20: Thinking about this some more... what is your thought on how users of zcrypto should set `LegacyNameString`?  Since it is a global, maybe it is safer to not export it and have a Get/Set function for it that uses a RWMutex.  Although since it is global I guess it really only makes since to set it once during initialization.

aspacewalz: It does seem like an odd thing to change dynamically during execution. 

I think the idiomatic path here is to leave concurrency to the consumer. If they are going to do something that dynamically changes this across routines, the consumer should be responsible for the locking.

aspacewalz (approved):

x509/pkix/pkix.go
dissoupov: in censys, we initialize it in the app once. there is no need for mutex or get/set.
either in `main()` or in `init() {
asn1.AllowPermissiveParsing = true
pkix.LegacyNameString = true
}`

codyprime: @dissoupov I just want to make sure we are not designing for a single downstream user's use case, since as an open-source project there may be many other users that pull in this functionality.  Drew's response makes sense though, so I am fine leaving it as-is.

codyprime (approved):

x509/pkix/pkix.go
dissoupov: this is not intended to be changed at run time.
by default, there will be standard functionality as in Go x509.
for apps that need different behavior, it should be initialized only once.

codyprime (approved):
LGTM, just had to re-read a section

PR #315, opened 2021-09-06
TLS 1.3: Add handshake tests

dissoupov:


tls/zcrypto_handshake_test.go
```diff
@@ -0,0 +1,120 @@
...
+		version uint16
+		addr    string
+	}{
+		// TODO SSL 3
+		//{VersionSSL30, "172.17.0.2:443"},
+		{VersionTLS10, "tls-v1-0.badssl.com:1010"},
+		{VersionTLS11, "tls-v1-1.badssl.com:1011"},
+		{VersionTLS12, "tls-v1-2.badssl.com:1012"},
+		// TODO: enable when supported
+		//{VersionTLS13, "tls-v1-2.badssl.com:1013"},
```
zakird, line 21: Is this the right name?

zakird (approved):

seglberg (approved):

tls/zcrypto_handshake_test.go
dissoupov: I tried `{VersionTLS13, "tls-v1-3.badssl.com:1013"}` and looks like they do not support TLS 1.3 yet.
But I updated the host now

PR #318, opened 2021-09-09
TLS1.3: added test for negociated Cipher Suite

dissoupov:


seglberg (approved):
I think @codyprime will have to approve as well to make it mergable.

tls/zcrypto_handshake_test.go
```diff
@@ -117,3 +119,106 @@ func TestCipherSuitesBadSSL(t *testing.T) {
...
+func TestCipherSuitesLocalTLS13(t *testing.T) {
+	server := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte(`OK`))
+	}))
+	defer server.Close()
+
+	tests := []struct {
+		cipherSuite    uint16
+		expectedSuites []uint16
+		err            string
```
dadrian, line 132: Doesn't look like this is ever non-empty

dadrian (approved):

tls/zcrypto_handshake_test.go
dissoupov: this is a placeholder. as we do expect there should be some errors when we start adding more ciphers from SSL3 etc

Issue #321, opened 2021-11-30
Can you add a field mean certificate has expired


xuemingjiang:
Can you add a field mean certificate has expired？

xuemingjiang:
#322 I have submitted a code

PR #322, opened 2021-11-30
Add certificate expiration field (#321)

xuemingjiang:
Add certificate expiration field

### DIFF
--- a/x509/json.go
+++ b/x509/json.go
@@ -212,6 +212,7 @@ type auxValidity struct {
 	Start          string `json:"start"`
 	End            string `json:"end"`
 	ValidityPeriod int    `json:"length"`
+	Expired        bool   `json:"expired"`
 }
 
 func (v *validity) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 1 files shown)


x509/json.go
```diff
@@ -212,13 +212,15 @@ type auxValidity struct {
...
 	ValidityPeriod int    `json:"length"`
+	Expired        bool   `json:"expired"`
 }
 
 func (v *validity) MarshalJSON() ([]byte, error) {
 	aux := auxValidity{
 		Start:          clampTime(v.NotBefore.UTC()).Format(time.RFC3339),
 		End:            clampTime(v.NotAfter.UTC()).Format(time.RFC3339),
 		ValidityPeriod: int(v.NotAfter.Unix() - v.NotBefore.Unix()),
+		Expired: (v.NotBefore.Unix() - time.Now().Unix()) > 0 || (time.Now().Unix() - v.NotAfter.Unix()) > 0,
```
dadrian, line 223: This should really be
```
now := time.Now()
expired := !v.NotBefore.Before(now) && !v.NotAfter.After(now)
```

That was we are clearly checking that it is not after `NotAfter` and not before `NotBefore`, and using the same time in both.

That being said, it's not clear to me we want this output to be time-dependent in the first place (i.e that we want this field).

dadrian (changes requested):

dadrian:
Closing due to lack of response.

PR #326, opened 2021-12-16
TLS 1.3: don't create skx log if processing server key exchange failed

codyprime:
If we attempt to create the skx log when `processServerKeyExchange` fails, then we will likely panic when looking up the elliptic curve.


seglberg (approved):

elliot-censys:
lgtm

PR #331, opened 2022-04-07
Added extension identifiers for Server Hello messages to Handshake Log

nakulbajaj:
In order to enable generating JA3S fingerprints for a TLS connection, we would like to include *all* extension identifiers as integers and include this information in TLS handshake logs.

### DIFF
--- a/tls/handshake_messages.go
+++ b/tls/handshake_messages.go
@@ -610,6 +610,7 @@ type serverHelloMsg struct {
 	selectedIdentityPresent      bool
 	selectedIdentity             uint16
 	supportedPoints              []uint8
+	extensionIdentifiers         []uint16
 
 	// HelloRetryRequest extensions
 	cookie        []byte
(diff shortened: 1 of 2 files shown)


zakird (approved):

zakird:
Code looks reasonable, can you look at why tests are failing?

nakulbajaj:
The MarshalUnmarshal test was failing - which seems to be because extension IDs are recorded when unmarshaling from bytes but not written to bytes when marshaling. New edits attempt to fix that by including empty data for all other extensions when marshaling.

tls/handshake_messages.go
```diff
@@ -636,17 +637,22 @@ func (m *serverHelloMsg) marshal() []byte {
 		var extensionsPresent bool
 		bWithoutExtensions := *b
 
+		var addedExtensions map[uint16]bool
+
```
codyprime, line 641: This map needs to be created, not just declared. e.g.:

```go
addedExtensions := make(map[uint16]bool)
```

nakulbajaj:
Since `MarshalUnmarshal` testing requires that the `serverHelloMsg` know its own extensions as it is initialized, we chose instead to create a function `extractExtensions` that will always give a slice of extension IDs calculated based on the raw bytes of the packet. This is the least messy solution that does not require modifying `serverHelloMsg` objects in either the `marshal` or `unmarshal` functions.

PR #333, opened 2022-04-21
Updated TLS_ECDHE_PSK_WITH_AES_128_GCM_SHA256 = 0xD001 = 53249

nakulbajaj:
It seems as though a typo was made when updating the TLS cipher suite mappings in Zcrypto. We would like to replace the old byte encoding for TLS_ECDHE_PSK_WITH_AES_128_GCM_SHA256 from 0xCAFE to 0xD001, as recorded by the [IANA](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4).


codyprime (approved):
Thanks @nakulbajaj, nice catch!  Change looks good to me, and thanks for keeping it ordered.

PR #344, opened 2022-08-16
Add option for CT log client to emit unparseable certs

elliotcubit:
- Allow CT log client to provide `IgnoreParsingErrors`, causing certs to be emitted regardless of parse status
- Standardize & export parsing error caused by unsupported elliptic curve


dadrian (approved):
Noting that you should look internally at csync code in pipeline to ensure this is actually used at Censys :)

PR #345, opened 2022-09-22
tls: perform type assertion on cert publickey

codyprime:
Prior to the TLS 1.3 backport, there was a type assertion to make sure that `cert.PublicKey.(*rsa.PublicKey)` was true.  This was lost in the backport work, and while very rare we did recently hit a case where this assertion is not true.  Doing it inline in the call leads to a panic.

This restores the prior type assertion check, and returns err if it fails.


elliotcubit (approved):
Looks good to me; thank you!

PR #348, opened 2023-01-31
Expose 'validSignature' field

elliotcubit:
Exposes `(*x509.Certificate).validSignature` so that it can be read without marshalling the entire certificate to JSON.


codyprime (approved):
Looks good to me; gofmt issues flagged by CI are not from this PR

PR #354, opened 2023-02-28
Create the various helper functions to enable complete Revocationlist support

aaomidi:


aaomidi:
@christopher-henderson just to make sure I understand, are you waiting on me to create a new PR introducing a CRL lint using this change, or are you planning on reviewing this PR first?

dadrian:
I'm fairly certain we can't take MPL code in this project, at least not in this directory.

dadrian:
If it's just the tests that are from Boulder, we should be able to:
- Convert them to table-driven tests
- Add the table (data) to a subfolder of `testdata` and slap MPL on just that folder

Can you clarify what part is MPL, and either remove it or provide an alternate information?

aaomidi:
> Can you clarify what part is MPL, and either remove it or provide an alternate implementation?

I think the main problem here is that we (zcrypto) doesn't have a pathway to merge from Go upstream and maintain the various custom changes that have been made. 

I believe I've made a note of everywhere this change differs from upstream. I'm open to suggestions on how to handle the few places the code differs. I think we can arguably get rid of it and see if we actually end up needing them.

dadrian (approved):

dadrian:
Thanks!

PR #358, opened 2023-04-13
verifier: add AppendFromPEMErr method

elliotcubit:
Return errors from the `io.Reader` here. Also return a slice of errors encountered while parsing certificates in the PEM.

Previously, it was necessary for the caller to know how many certificates were in the input PEM in order to check this function for error.


verifier/graph.go
```diff
@@ -254,19 +254,23 @@ func (g *Graph) IsRoot(c *x509.Certificate) bool {
...
 
 // AppendFromPEM adds any certificates encoded as PEM from r to the graph. If
-// root is true, it marks them as roots. It returns the number of certificates
-// parsed.
-func (g *Graph) AppendFromPEM(r io.Reader, root bool) int {
+// root is true, it marks them as roots. It returns:
+// - The number of certificates parsed
+// - A slice of errors encountered while parsing certificates
+// - Any non-EOF error encountered reading from r
+func (g *Graph) AppendFromPEM(r io.Reader, root bool) (int, []error, error) {
```
codyprime, line 261: To maintain API backwards compatibility, how about renaming this func as you have it written, to `AppendFromPEMReportErr` (or something like that), and then have `AppendFromPEM` wrap that function, but ignore the error returns (i.e., keep the original function signature).  That way existing consumers of the API outside this project won't it an API change.

codyprime (changes requested):

codyprime:
The CI failure is unrelated to this PR, and is due to go version (1.20.3 being use by CI).  The TLS 1.3 Feature branch needs to be updated, but that doesn't need to block this PR.

codyprime (approved):
The CI failure is unrelated to this PR, and is due to go version (1.20.3 being use by CI). The TLS 1.3 Feature branch needs to be updated, but that doesn't need to block this PR.

PR #360, opened 2023-04-28
verifier: set ValidSignature for certificates in the graph

elliotcubit:
Certificates which are used to make trust decisions (present in the validation graph) did not correctly have the `ValidSignature` field set when they were verified.

The signatures of certificates in the graph _are_ valid, so change this.


dadrian (approved):

dadrian:
You sure you don't want this in the `main` branch?

elliotcubit:
> You sure you don't want this in the `main` branch?

Yes, this field hasn't hit `main`, so the patch isn't needed.

PR #363, opened 2023-08-29
ct/x509: fix dropped error

alrs:
This fixes a dropped `err` variable in `ct/x509`.

### DIFF
--- a/ct/x509/x509.go
+++ b/ct/x509/x509.go
@@ -66,6 +66,9 @@ func marshalPublicKey(pub interface{}) (publicKeyBytes []byte, publicKeyAlgorith
 			N: pub.N,
 			E: pub.E,
 		})
+		if err != nil {
+			return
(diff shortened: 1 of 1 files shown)


ct/x509/x509.go
```diff
@@ -66,6 +66,9 @@ func marshalPublicKey(pub interface{}) (publicKeyBytes []byte, publicKeyAlgorith
 			N: pub.N,
 			E: pub.E,
 		})
+		if err != nil {
+			return
```
zakird, line 70: Should we return the err itself since we have a handle to it?

alrs: `err` is defined in the function signature, so `return` is enough for `err` to be returned.

https://go.dev/play/p/VnE-UfpEdDE

zakird (approved):

PR #372, opened 2023-10-24
Make ocsp stapling and supported EC points configurable

esuwu:
I left the default behavior of ClientHello the same but made some fields configurable through Config.


esuwu:
@zakird could you please take a look at this PR?

zakird (approved):

Issue #373, opened 2023-10-24
Add support for subject:emailAddress field when marshalling x509 certificates


robplee:
There's currently an effort to add lints to zlint to test for the various rules from the new (ish) SMIME BR certificates.  Many of these lints will be checking the subject.EmailAddress field however creating test data is a bit of a pain because the zcrypto/x509.CreateCertificate function doesn't include anything specified in pkix.Name-EmailAddress fields in the marshalled ASN.1.  It won't be a difficult fix, just adding another line to Name.ToRDNSequence() (x509/pkix/pkix.go#268).

I'm happy to make the change+PR but I thought it'd be worth creating an issue about this first before opening the PR in case anyone has a firm objection to this whole idea!

zakird:
No objections. Makes sense to me

PR #375, opened 2023-11-02
Add email addresses to ASN.1 marshalled Name RDN sequences

robplee:
Very small PR to address #373 which should make it easier to make test certificates using the genTestCerts.go script in zlint.


robplee:
@zakird, you mentioned on the issue (#373) you have no objections to the change and that it made sense to you - could I trouble you for a quick review?

zakird (approved):

Issue #381, opened 2024-02-24
Remove the message fragmentation mechanism in TLS 1.0


szy1231:
In considering security, the Go official implementation includes a message fragmentation mechanism in the TLS 1.0 protocol. This design decision results in the inability to obtain complete information when exchanging NTLM information using the TLS 1.0 version. Given that the primary use of this library is for data collection, it is recommended to adjust the TLS 1.0 implementation to remove the fragmentation restriction.

Thank you for your assistance.

PR #382, opened 2024-02-24
remove the message fragmentation mechanism in TLS 1.0

szy1231:
Remove the message fragmentation mechanism in TLS 1.0 #381


zakird (approved):

zakird:
Overall, this makes sense to me, but it does look like tests now fail with the change. Can you take a look?

PR #387, opened 2024-08-28
make extension parsing more permissive

iJustErikk:
if allowpermissive, then tolerate errors in extension parsing

searched for returns in parsecertificate. continued in each case for an easy check that this is as nil safe as the code was before

here is a cert that we can now permissively parse:

`MIICQDCCAamgAwIBAgIQa9ciWy1lvZtGRujXx93btjANBgkqhkiG9w0BAQUFADBAMRowGAYDVQQDExFEaWdpQ2VydCBzZWN1cml0eTEVMBMGA1UEChMMRGlnaUNlcnQgSW5jMQswCQYDVQQGEwJVUzAeFw0yMzA3MDQyMTE3MjRaFw0yNTA3MDQyMTE3MjRaMEAxGjAYBgNVBAMTEURpZ2lDZXJ0IHNlY3VyaXR5MRUwEwYDVQQKEwxEaWdpQ2VydCBJbmMxCzAJBgNVBAYTAlVTMIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQCoqY2vi90BoFCGapgvjh2uCtqKBOFjN+3kU2Lc2UuSlt/pbJIBsRK+pAFkgy1khodbkzthiuzzDzqb95Y9a3eJ+G3dx1aW872v8ub8qe/ZlXOVDnNh/AahIzKlBGE3cuCT7eHJnGHt2tUmAoCc5OodgXdso6vKYxdYrszLrDnYWQIDAQABozswOTALBgNVHQ4EBAMCA7gwCwYDVR0PBAQDAgO4MB0GA1UdJQQWMBQGCCsGAQUFBwMBBggrBgEFBQcDAjANBgkqhkiG9w0BAQUFAAOBgQCnChlYocxRYGl7QSq5luUpak2L5u0PcsEzCUd9yIUCB1eubqMVQTJfp7frMWM313aBpNeWQ3leLZvNn3hzPPBWy8ZH18plZRjkTxh9U7cEK3I/yltzLlZdUPNqgEJdeHMHYjKtO7LcZATm+DAhkk0HdX0KS4dN0G+zFKFWCEPhCg==`

### DIFF
--- a/tls/handshake_client.go
+++ b/tls/handshake_client.go
@@ -20,6 +20,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/zmap/zcrypto/encoding/asn1"
 	"github.com/zmap/zcrypto/x509"
 )
 
@@ -853,7 +854,7 @@ func (c *Conn) verifyServerCertificate(certificates [][]byte) error {
 	certs := make([]*x509.Certificate, len(certificates))
 	for i, asn1Data := range certificates {
 		cert, err := x509.ParseCertificate(asn1Data)
-		if err != nil {
+		if err != nil && (!asn1.AllowPermissiveParsing || !strings.HasPrefix(err.Error(), "permissive")) {
 			c.sendAlert(alertBadCertificate)
 			return errors.New("tls: failed to parse certificate from server: " + err.Error())
 		}
--- a/x509/x509.go
+++ b/x509/x509.go
@@ -1668,6 +1668,9 @@ func parseCertificate(in *certificate) (*Certificate, error) {
 	out.SubjectUniqueId = in.TBSCertificate.SubjectUniqueId
 
 	out.ExtensionsMap = make(map[string]pkix.Extension, len(in.TBSCertificate.Extensions))
+
+	// collect all permissive parse errors for extensions
+	var permissiveErrors []error
 	for _, e := range in.TBSCertificate.Extensions {
 		out.Extensions = append(out.Extensions, e)
 		out.ExtensionsMap[e.Id.String()] = e
@@ -1706,6 +1709,10 @@ func parseCertificate(in *certificate) (*Certificate, error) {
 					out.URIs, out.DirectoryNames, out.EDIPartyNames,
 					out.IPAddresses, out.RegisteredIDs, out.FailedToParseNames, err = parseGeneralNames(e.Value)
 				if err != nil {
+					if asn1.AllowPermissiveParsing {
+						permissiveErrors = append(permissiveErrors, err)
+						continue
+					}
 					return nil, err
 				}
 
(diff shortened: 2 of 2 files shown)


iJustErikk:
yeah I could just have an error that we'll conditionally toss permissive errors into and prefix it with `permissive: `

zakird:
Yeah, I think that makes sense.

codyprime:
> yeah I could just have an error that we'll conditionally toss permissive errors into and prefix it with permissive: . Does that sound good?

Will this still work as the original PR intended? With this error appending addition to the PR, `ParseCertificate` will now return a non-nil error even with `asn1.AllowPermissiveParsing`.

iJustErikk:
> > yeah I could just have an error that we'll conditionally toss permissive errors into and prefix it with permissive: . Does that sound good?
> 
> Will this still work as the original PR intended? With this error appending addition to the PR, `ParseCertificate` will now return a non-nil error even with `asn1.AllowPermissiveParsing`.

the user can check for "permissive" to know if the parsing only had permissive errors. additionally, we only append on the new checks, so users of the old "just ignore it" behavior won't be affected

codyprime (approved):
thanks, test failures look unrelated to this PR and are pre-existing on the `feature/TLSv1.3` branch

elliotcubit:
I missed the boat here, I think we should return without an error? That's the behavior everywhere else we do permissive parsing.

If we really want to start returning these errors, we should do so with an explicit `go` type, but I don't think we want that?

codyprime:
FWIW, I'm fine returning without an error, since it is explicitly requested with `AllowPermissiveParsing` to being with

elliotcubit:
This line in the patch handling the error for this one case was the red flag for me -- this pattern becomes necessary everywhere that permissive parsing is desired.

```
if err != nil && (!asn1.AllowPermissiveParsing || !strings.HasPrefix(err.Error(), "permissive")) {
```

Though it isn't bad in theory, it breaks the API of `AllowPermissiveParsing` flag.

zakird:
@codyprime  I guess I'm thinking for something like Censys, we'd probably want to allow permissive parsing, always. But we would also want to know that it failed?

@elliotcubit Yeah, agree on checking permissive is gross and we probably want a different way to express it back to the caller.

elliotcubit:
> But we would also want to know and mark in our data that it failed

It's already the case that fields might be missing with no indication otherwise; when the permissive parsing flag allows a malformed cert, Censys doesn't mark it as incomplete, because zcrypto doesn't indicate that.

iJustErikk:
sooo do we want to break the api for AllowPermissiveParsing or do we want to give back errors?

only other choice I can think of is keeping ParseCertificate and having a separate ParseCertificateWithPermissiveErrors with the new behavior

PR #388, opened 2024-10-25
ignore permissive errors

iJustErikk:
as per the discussion in https://github.com/zmap/zcrypto/pull/387, let's not break the previous allowpermissive api.


elliotcubit (approved):
Looks good to me. Thank you!

PR #428, opened 2025-06-24
handshake log: add a field for TLS alerts in hslog

scohen-censys:
### Description
Add an additional field to the handshake log for TLS alerts. The field contains the alert description as an int, and can be converted to the human-readable form using existing functions.

### How to Test
```
zcrypto$ go test ./...
```


tls/tls_handshake.go
```diff
@@ -147,6 +147,7 @@ type ServerHandshake struct {
 	SessionTicket      *SessionTicket     `json:"session_ticket,omitempty"`
 	ServerFinished     *Finished          `json:"server_finished,omitempty"`
 	KeyMaterial        *KeyMaterial       `json:"key_material,omitempty"`
+	Alert              *alert             `json:"alert,omitempty"`
```
codyprime, line 150: We probably want to also export `alert`

scohen-censys: We can get away with not exporting since it implements `error`, but if you think it makes more sense to export, I agree

codyprime (changes requested):
The test failures for the the alert name change appear to be legit, so we should update those

dadrian (approved):
:shipit:

dadrian:
👋 hi jeff sorry I didn't come back

codyprime (approved):

PR #449, opened 2025-10-28
Update deps and ignore < 1024 bit RSA key size errors in tests

phillip-stephens:
- Updates our `crypto` and `net` dependencies
   - The new crypto dependency adds runtime checks that prevent certificate operations on RSA keys of less than 1024 bits: https://pkg.go.dev/crypto/rsa#hdr-Minimum_key_size
- Enables pkg wide this `GODEBUG` flag to disable the check. Tested by updating the ZGrab2's zcrypto dependency to this branch and I can still find `512` bit RSA keys.

### DIFF
--- a/ct/x509/x509.go
+++ b/ct/x509/x509.go
@@ -19,6 +19,7 @@ import (
 	"crypto/rsa"
 	"crypto/sha1"
 	"os"
+	"strings"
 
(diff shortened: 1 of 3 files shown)


zakird:
Wouldn't we want to disable this for any users of ZCrypto? We want to be able to successfully handshake with everything on the Internet regardless of key size.

PR #487, opened 2026-04-11
tls: respect ForceSuites in makeClientHello and add test

### LINKED ISSUE #485: GREASE and other meta cipher suites not included in client hellos with ForceSuites
I've been experimenting with porting some of the test code in howsmyssl.com (repo: https://github.com/jmhodges/howsmyssl) to zcrypto/tls.

For context, howsmyssl is a project that tells folks about the status of their web client's TLS code. zcrypto/tls is a good fit for making sure it handles older cipher suites and protocols even as the Go stdlib drops support for broken things.

However, I get test failures because, even with `ForceSuites: true` on the config, zcrypto/tls prevents any GREASE and renegotiation cipher suites from being in the client hello. I'm sure I'm just holding the library wrong, but I'm not sure what boolean I've missed. 

(The specific test is for detecting Sweet32 mitigations where you put the Sweet32 cipher suites after all the other cipher suites but before any meta cipher suites like GREASE and renegotiation).

I've got a repro below. From a cursory look, it seems like zcrypto/tls checks `ForceSuites` in `ClientFingerprintConfiguration.marshal`, but doesn't in `Conn.makeClientHello`. But I'm not sure if that's expected for a bug

While the renegotiation cipher suite might be added a different way, the example GREASE cipher suite I included is not (nor are any other newer meta cipher suites that get added).

Here's what I've got for a repro:

```go
func TestForceSuitesCipherSuiteOrdering(t *testing.T) {
	c, s := localPipe(t)
	defer s.Close()

	expectedSuites := []uint16{
		TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
		TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA,
		0x0A0A, // a GREASE cipher suite
		0x00FF, // renegotiation
	}

	go func() {
		client := Client(c, &Config{
			InsecureSkipVerify: true,
			CipherSuites:       expectedSuites,
			ForceSuites:        true,
			MaxVersion:         VersionTLS12,
		})
		client.Handshake()
		c.Close()
	}()

	var header [5]byte
	if _, err := io.ReadFull(s, header[:]); err != nil {
		t.Fatal(err)
	}
	recordLen := int(header[3])<<8 | int(header[4])

	record := make([]byte, recordLen)
	if _, err := io.ReadFull(s, record); err != nil {
		t.Fatal(err)
	}

	var m clientHelloMsg
	if !m.unmarshal(record) {
		t.Fatal("failed to unmarshal ClientHello")
	}

	if !reflect.DeepEqual(m.cipherSuites, expectedSuites) {
		t.Fatalf("cipher suites in ClientHello = %#v, want %#v", m.cipherSuites, expectedSuites)
	}
}
```

jmhodges:
ForceSuites was only honored in the ClientFingerprintConfiguration
marshal path, not in makeClientHello used by Client().Handshake().
Unknown cipher suites were silently dropped. Now when ForceSuites is
true, suites not in the internal cipherSuites list are still included
in the ClientHello.

Fixes #485

### DIFF
--- a/tls/handshake_client.go
+++ b/tls/handshake_client.go
@@ -275,19 +275,23 @@ func (c *Conn) makeClientHello() (*clientHelloMsg, map[CurveID]tls13KeyShare, er
 	possibleCipherSuites := config.cipherSuites()
 	hello.cipherSuites = make([]uint16, 0, len(possibleCipherSuites))
 
-	for _, suiteId := range possibleCipherSuites {
-		for _, suite := range cipherSuites {
-			if suite.id != suiteId {
-				continue
-			}
-			// Don't advertise TLS 1.2-only cipher suites unless
-			// we're attempting TLS 1.2.
-			if hello.vers < VersionTLS12 && suite.flags&suiteTLS12 != 0 {
+	if !config.ForceSuites {
+		for _, suiteId := range possibleCipherSuites {
+			for _, suite := range cipherSuites {
+				if suite.id != suiteId {
+					continue
+				}
+				// Don't advertise TLS 1.2-only cipher suites unless
+				// we're attempting TLS 1.2.
+				if hello.vers < VersionTLS12 && suite.flags&suiteTLS12 != 0 {
+					break
+				}
+				hello.cipherSuites = append(hello.cipherSuites, suiteId)
 				break
 			}
-			hello.cipherSuites = append(hello.cipherSuites, suiteId)
-			break
 		}
+	} else {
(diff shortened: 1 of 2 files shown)


zakird (approved):

jmhodges:
@zakird Thanks! Does this approval mean it should merge in? (If so, I don't have the write access to do so)

zakird:
Gah sorry. 🤦🏽  merging!

PR #498, opened 2026-05-13
Add a fork of RSA to allow us to parse large exponents in TLS certs and SSH handshakes

phillip-stephens:
In looking into [zgrab issue #670](https://github.com/zmap/zgrab2/issues/670), I realized we'd need to fork `crypto/rsa` to add proper support. While the SSH handshake could be handled within the ssh fork in `zgrab2`, handling the TLS case isn't possible without changes to `zcrypto`.

In order to attempt to make this easier to maintain, I've done the following:
- Added the tagged release branch I copied from the stdlib to `rsa/FORK_SOURCE.md` so it's clear for future readers where this came from
- Annotated where this version diverged with comments - `// ZCrypto - explanation of change`
- Switched all uses of `crypto/rsa` within `GitHub.com/zmap/zcrypto` to use the fork.

Changes Summary for the RSA fork
- Removed any references to any `internal/` packages that won't compile outside of the stdlib. Fortunately weren't many
- Switched to using `big.Int` for the exponent and the relevant operations on `E` had to be swapped to the `big.Int` versions
- Commented out any code that depends on `boring`/BoringSSL since we don't use it in ZCrypto

Testing
- [x] Perform a large scale A/B test with the TLS scanner in ZGrab using this fork, ensure the results are identical
- [x] Perform a large scale A/B test with the SSH scanner in ZGrab using this fork, ensure the results are identical except for the few hosts that had a large exponent (which ofc should now be able to be parsed)


zakird (approved):
Looks good. Understand that you want to do some more testing before you merge. Let me know if you make any significant changes and want a pair of eyes on it again.

Issue #514, opened 2026-07-15
Fix PortableCiphers so that they're portable instead of un-portable on the modern Interweb


mzpqnxow:
Hello Zmap folks,

I did some work a long time ago to make `zgrab2` more robust in its TLS negotiation. The changes for `zcrypto` are in #238, which just adds a group of ciphers to a special list in `zcrypto`, called `tls.PortableCiphers`.

Some of this is documented or related to #236 , #229 , #342, though I promise it's not too interesting. It was reasonably useful at the time, as the default cipher-suite was not as complete as it could have been. It resulted in quite a few less failed handshakes.

That was before the TLSv1.3 branch was done or merged, if I recall.

Fast forward to the modern era. To my surprise, `echo google.com | zgrab2 http --cipher-suite portable` fails to negotiate (though default settings negotiate fine - oops, that's un-portable!)

I took a quick look why and noticed that I omitted the TLSv1.3 ciphers. Easy to fix.

tl; dr This issue is a placeholder for a PR that fixes my own mess - `PortableCiphers` became un-portable, because TLS1.3 ciphers weren't available at the time, or I was careless (or both)

PR just adds the TLSv1.3 ciphers and will arrive shortly.

Thanks!

PR #515, opened 2026-07-15
Update the tls portable ciphers list

mzpqnxow:
This is for #514  

Note, as I believe was mentioned a while ago, the other collections of cipher-suites in this file (e.g. Firefox, Safari, Chrome, etc) are very possibly/probably not up to date.

I _think_ the consensus was that was OK - they're best-effort, and not necessarily relied upon by anyone who is making noise.

Regardless, I don't have the time to look into whether they're correct, otherwise I would address them. This change is just to make the portable list complete.

Thanks!


mzpqnxow:
To be clear - this change will only impact those who explicitly specify the use of `tls.PortableCiphers` - which is probably not many people, or someone would have complained about this issue before me :laughing:

zakird (approved):

PR #519, opened 2026-07-22
Add post-quantum key-exchange group MLKEM1024

mbotero853:
Finishing up https://github.com/zmap/zcrypto/pull/516, adds support for ML-KEM-1024, which is the strongest standardized pure PQC group supported by TLS 1.3.

Again, this was tested as both a client and server, including connecting to a real website that supports this group.


zakird:
This seems reasonable, but it's difficult to know without any tests. Any chance you'd able to add tests similar to what's in place for other key exchanges?

mbotero853:
> This seems reasonable, but it's difficult to know without any tests. Any chance you'd able to add tests similar to what's in place for other key exchanges?

True, good idea. I've added client tests for all 4 post-quantum key-exchange groups.

I had to update our OpenSSL version, and that's a big enough change that I decided to split it into a separate PR: https://github.com/zmap/zcrypto/pull/521 . I cherry-picked that PR into this one, so the other PR should probably be merged first.

Also, I updated `defaultCurvePreferences` to include the new curves (which I hadn't done earlier because I was copying the original PR that added a post-quantum curve, but now I realize it's probably a good idea) but this changed some of the flows, so some irrelevant testdata had to be regenerated too. I verified that without the `defaultCurvePreferences` update the old tests do still pass without changes, though.

I also had to change a couple things with zcrypto/tls's usage of MLKEM to not use the default random -- namely replacing `ek.Encapsulate` and `mlkem.GenerateKey1024` with appropriate alternatives.

zakird (approved):

PR #528, opened 2026-08-27
Pqc support

matthewzchou:
Finished Pure MLKEM 512, 768, and 1024 cases for tls and added MLDSA to tls directory.
Test Cases still need to be updated as I am unable to compile them on my OS.
MLDSA and MLKEM are currently supported but unadvertised by default


zakird:
Code looks reasonable. Why all of the changes to the existing unrelated test certs?

matthewzchou:
The problem was that the signature algorithms weren't accounted for in the hard coded test cases which caused them to be errors. I fixed it though and I am about to add a new PR.

tls/auth.go
```diff
@@ -17,6 +17,10 @@ import (
 
 	rsa "github.com/zmap/zcrypto/rsa"
 	"github.com/zmap/zcrypto/x509"
+
+	"github.com/cloudflare/circl/sign/mldsa/mldsa44"
+	"github.com/cloudflare/circl/sign/mldsa/mldsa65"
+	"github.com/cloudflare/circl/sign/mldsa/mldsa87"
```
dadrian: Can we bump go versions instead and use the version in the standard library?

tls/handshake_client.go
```diff
@@ -874,7 +891,11 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		hs.finishedHash.Write(skx.marshal())
 		err = keyAgreement.processServerKeyExchange(c.config, hs.hello, hs.serverHello, c.peerCertificates[0], skx)
 		if err != nil {
-			c.sendAlert(AlertUnexpectedMessage)
+			if errors.Is(err, errMLDSAInTLS12) {
+				c.sendAlert(AlertIllegalParameter)
```
dadrian: Not convinced we want this behavior since ZCrypto is generally permissive for measurement purposes. We could probably complete a handshake with ML-DSA signatures in TLS 1.2, even if it's nominally against the spec, since I don't think any structural changes are needed to the messages. Whereas ML-KEM key exchange is definitely not happening in TLS 1.2.

tls/handshake_extensions.go
```diff
@@ -304,22 +303,37 @@ type SignatureAlgorithmExtension struct {
...
 
 func (e *SignatureAlgorithmExtension) WriteToConfig(c *Config) error {
-	c.SignatureAndHashes = e.getStructuredAlgorithms()
+	algs := e.getStructuredAlgorithms()
+	c.SignatureAndHashes = make([]SigAndHash, 0, len(algs))
+
+	for i, alg := range algs {
+		switch SignatureScheme(e.SignatureAndHashes[i]) {
+		case MLDSA44Sig, MLDSA65Sig, MLDSA87Sig:
+			continue
```
dadrian: Don't we want these to be included?

tls/handshake_server_tls13.go
```diff
@@ -590,7 +590,7 @@ func (hs *serverHandshakeStateTLS13) sendServerCertificate() error {
 		certReq := new(certificateRequestMsgTLS13)
 		certReq.ocspStapling = true
 		certReq.scts = true
-		certReq.supportedSignatureAlgorithms = supportedSignatureAlgorithms
+		certReq.supportedSignatureAlgorithms = supportedSignatureAlgorithmsTLS13
```
dadrian: Again, we could probably still do the right thing in TLS 1.2 since ZCrypto is specifically a permissive library for measurement.

tls/key_agreement.go
```diff
@@ -128,6 +129,10 @@ func (ka *signedKeyAgreement) verifyParameters(config *Config, clientHello *clie
 		if len(sig) < 2 {
 			return nil, errServerKeyExchange
 		}
+		scheme := SignatureScheme(sigAndHash[0])<<8 | SignatureScheme(sigAndHash[1])
+		if isTLS13OnlySignatureAlgorithm(scheme) {
+			return nil, errMLDSAInTLS12
+		}
```
dadrian: (same comment)

tls/key_schedule.go
```diff
@@ -434,6 +437,56 @@ func (k *tls13SecP384r1MLKEM1024) SharedKey(serverShare []byte) ([]byte, error)
 	return slices.Concat(ecdheSS, kemSS), nil
 }
 
+// Using CIRCL's MLKEM512 implementation
```
dadrian, line 440: I'll poke Filippo about getting MLKEM512 into Go.

dadrian (commented):
Some initial thoughts

dadrian:
Thanks for implementing this!

tls/handshake_extensions.go
matthewzchou: Yeah, my mistake, I was trying to reject cases where MLDSA was in TLS 1.2 since that was RFC specification

tls/handshake_client.go
matthewzchou: Fixed it

tls/auth.go
matthewzchou: Yup, I'm working on it

matthewzchou:
I updated everything, however, this requires go to be in 1.27

zakird:
Can you update tests to use Go 1.27 then? We can't merge this with tests failing.

matthewzchou:
Should be updated now

